PR #32, opened 2024-09-16
ci: adapt release workflows

ccharly:
## Description

Current workflows files were implemented for single packages.

This PR imports the monorepo release logic from the [`core`](https://github.com/MetaMask/core) monorepo.


.github/workflows/main.yml
```diff
@@ -51,28 +51,25 @@ jobs:
...
     outputs:
       IS_RELEASE: ${{ steps.is-release.outputs.IS_RELEASE }}
-    runs-on: ubuntu-latest
     steps:
-      - uses: MetaMask/action-is-release@v1
-        id: is-release
+      - id: is-release
+        uses: MetaMask/action-is-release@v2
+        with:
+          commit-starts-with: '[version]'
```
Mrtenz, line 64: I don't think this works, unless you want the commits to be "x.y.z" literally (without "Release" in front).

ccharly: Actually, that's what I wanted, to match the other repos we had like: https://github.com/MetaMask/eth-snap-keyring/commit/a242534fe021a58ba7d244e48d81fe1c9d958f5f

But.. since those commits are already part of the history, this might look weird indeed 🤔

ccharly: Finally I went for `release: [version]`, I took inspiration from angular repo

Mrtenz: Sounds good. Make sure to document it somewhere to avoid confusion.

danroc (approved):

PR #40, opened 2024-09-18
build!: build packages with `ts-bridge`

Mrtenz:
This updates all packages to build with `ts-bridge` instead of `tsc`. This has the benefit that packages will be built as CJS modules and ESM modules at the same time.

I've also modernised `eth-hd-keyring` in this process, as it was previously a single JS file. For now it's running TypeScript with `allowJs`, but this file can be properly converted to TypeScript in the future.

## Breaking changes

- Each package now has an `exports` field, meaning subpath imports (`import '@metamask/eth-hd-keyring/dist/foo';`) are no longer possible, unless explicitly stated in the exports field.


legobeat:
Might want to couple this with an upgrade of `typescript` to min 5.5: https://github.com/ts-bridge/ts-bridge/issues/67#issuecomment-2421465052

ccharly:
@metamaskbot publish-previews

ccharly:
I was able to run the E2E tests on the extension side using a bunch of previews builds, everything looks fine.

Some tests are failing, but I truly think this is not related to that PR, but more about the latest changes we made on the mainline (the extension + the `core` repo are not up-to-date with our latest release).

Here's the PRs I used for tests:
- https://github.com/MetaMask/metamask-extension/pull/28455
- https://github.com/MetaMask/core/pull/4932

Here's the CI report:
Screenshot 2024-11-15 at 10 13 50
Screenshot 2024-11-15 at 10 23 08

- The `dedupe` fail being a non-issue here
- The 2 failing tests might be because of latest changes, but not related to the move to `ts-bridge`

ccharly (approved):

PR #43, opened 2024-09-19
ci: update actions, dependabot config, and internal dependencies

danroc:
The current Dependabot configuration is not working. This change copies the configuration from [MetaMask/core](https://github.com/MetaMask/core) to attempt to fix the issue.

It also updates some actions and adds the `--immutable-cache` flag to some `yarn install` invocations.

The `@metamask/keyring-api` dependency used by `@metamask/eth-snap-keyring` is moved to `peerDependencies` and bumped to 8.1.1.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -18,14 +18,8 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
-      - name: Install Yarn
-        run: corepack enable
-      - name: Restore Yarn cache
-        uses: actions/setup-node@v4
-        with:
-          node-version: ${{ matrix.node-version }}
           cache: yarn
-      - run: yarn --immutable
+      - run: yarn install --immutable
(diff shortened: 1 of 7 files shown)


.github/workflows/build-lint-test.yml
```diff
@@ -41,13 +35,13 @@ jobs:
...
+      - uses: actions/checkout@v4
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
-          cache: 'yarn'
-      - run: yarn --immutable
+          cache: yarn
+      - run: yarn install --immutable --immutable-cache
```
ccharly, line 44: Just for the record, I did remove it because of some caching issue, but maybe the underlying problem was not the cache itself, but more the dependencies!

- https://github.com/MetaMask/accounts/actions/runs/10900174593/job/30247230364

But let's try to re-enable it for the moment 👍

ccharly (approved):

PR #47, opened 2024-09-20
feat: add `yarn audit` step

ccharly:
Adds a `yarn audit` step to try to catch any "vulnerable" dependencies.


danroc:
Maybe we should remove the `--recursive` flag since we have no control over transaitive dependencies.

PR #48, opened 2024-09-20
docs: update how-to-release.md

ccharly:
The command-line was wrong:
- `\n` was not interpreted by `gh`
- `VERSION=x.y.z <command>` is not being interpreted by the `<command>` and cannot be expanded, thus adding a `;` to first declare the variable and then use it


ccharly:
Dropping in favor of:
- https://github.com/MetaMask/accounts/pull/49

PR #55, opened 2024-09-24
chore: update lint tools versions + use ts 5

ccharly:
Updating `eslint` and other similar packages.


packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -109,15 +115,15 @@ export class LedgerKeyring extends EventEmitter {
...
     return this.bridge.init();
   }
 
-  async destroy() {
+  async destroy(): Promise<void> {
     return this.bridge.destroy();
   }
 
-  async serialize() {
+  async serialize(): Promise<Partial<LedgerBridgeKeyringOptions>> {
```
ccharly, line 126: I did this, but maybe we should have a dedicated type for the state here?

packages/keyring-eth-trezor/src/trezor-connect-bridge.ts
```diff
@@ -35,29 +40,36 @@ export class TrezorConnectBridge implements TrezorBridge {
     this.trezorConnectInitiated = true;
   }
 
-  dispose() {
+  async dispose(): Promise<void> {
```
ccharly, line 43: Was not `async` but since it's returning a `Promise` it makes sense to make it `async`

packages/keyring-eth-trezor/src/trezor-keyring.ts
```diff
@@ -116,14 +125,14 @@ export class TrezorKeyring extends EventEmitter {
     return this.bridge.model;
   }
 
-  init() {
+  async init(): Promise<void> {
```
ccharly, line 128: Was not `async` but since it's returning a `Promise` it makes sense to make it `async`.

Also, the `KeyringController` is already `await`ing those calls: https://github.com/MetaMask/core/blob/main/packages/keyring-controller/src/KeyringController.ts#L2199-L2201

packages/keyring-eth-trezor/src/trezor-keyring.ts
```diff
@@ -443,15 +453,16 @@ export class TrezorKeyring extends EventEmitter {
     });
   }
 
-  /**
-   * EIP-712 Sign Typed Data
-   */
+  // EIP-712 Sign Typed Data
```
ccharly, line 456: Using a simple comment here, since most of those method are using this pattern

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -585,7 +589,7 @@ export class SnapKeyring extends EventEmitter {
    * @param snapId - The Snap ID to check allowed origins for.
    * @throws An error if the URL's origin is not in the Snap's allowed origins.
    */
-  #validateRedirectUrl(url: string, snapId: SnapId) {
+  async #validateRedirectUrl(url: string, snapId: SnapId): Promise<void> {
```
ccharly, line 592: Seems non-necessary, but this is a private method, so making it `async` is ok here.

ccharly (commented):
I did review (but cannot approve since I have some commits here). Left some comments with some open-questions + to explain some of the changes that were made by me.

jest.config.packages.js
```diff
@@ -141,7 +141,7 @@ module.exports = {
   // runner: "jest-runner",
 
   // The paths to modules that run some code to configure or set up the testing environment before each test
-  setupFiles: ['../../tests/setup.ts'],
+  // setupFiles: ['../../tests/setup.ts'],
```
gantunesr, line 144: Why comment it out?

packages/keyring-api/src/rpc-handler.ts
```diff
@@ -47,22 +47,22 @@ async function dispatchRequest(
   assert(request, JsonRpcRequestStruct);
 
   switch (request.method) {
-    case KeyringRpcMethod.ListAccounts: {
+    case `${KeyringRpcMethod.ListAccounts}`: {
```
gantunesr, line 50: I'm not following on why this change is necessary, from a first look it seems redundant

packages/keyring-eth-ledger-bridge/src/ledger-hw-app.ts
```diff
@@ -1,16 +1,8 @@
...
 import { Buffer } from 'buffer';
 
 import { GetAppNameAndVersionResponse } from './type';
 
-// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
-export interface MetaMaskLedgerHwAppEth extends LedgerHwAppEth {
-  openEthApp(): void;
-  closeApps(): void;
-  getAppNameAndVersion(): Promise<GetAppNameAndVersionResponse>;
-}
```
gantunesr, line 12: Just confirming that this is unused

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -36,6 +34,14 @@ type SignTransactionPayload = Awaited<
   ReturnType<LedgerBridge<LedgerIframeBridgeOptions>['deviceSignTransaction']>
 >;
 
+type AccountPageEntry = {
+  address: string;
+  balance: number | null;
+  index: number;
+};
+
+type AccountPage = AccountPageEntry[];
```
gantunesr, line 43: Should these be exported?

packages/keyring-eth-ledger-bridge/test/document.shim.ts
```diff
@@ -1,24 +1,31 @@
-// eslint-disable-next-line import/no-mutable-exports
+// eslint-disable-next-line import-x/no-mutable-exports
```
gantunesr, line 1: I'm not familiar with this one, how is it different from `import/no-mutable-exports`?

gantunesr (commented):
Looking good, left a few comments

packages/keyring-eth-ledger-bridge/src/ledger-hw-app.ts
ccharly: This is indeed unused and redundant here (while also be kinda invalid 🙈)

- The `interface` is using the same name than the `class` following it, meaning the `class` will override it anyway. I believe it has been introduced to force the methods signatures on the `class` itself
- `interface` can extends `class` to some extent but it seems to have some restrictions (specifically with private members)

packages/keyring-api/src/rpc-handler.ts
ccharly: `eslint was complaining here, most likely cause `request.method` is being infered as `string` (which makes sense given that `JsonRpcRequestStruct` is using `method: string()`.


For reference:
- https://typescript-eslint.io/rules/no-unsafe-enum-comparison/

I assume it's considered unsafe to compare a `string` with an `enum` value here, hence the use of template literal to go from (type-wise) `string === enum` to `string === string`

packages/keyring-eth-ledger-bridge/test/document.shim.ts
ccharly: Looks like the plugin got renamed:

- https://github.com/un-ts/eslint-plugin-import-x/issues/24#issuecomment-1991605123

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
ccharly: Neither `eslint` nor `tsc` complained about this type not being exported, but IMO we should since it's part of public functions signatures, lemme export them! Thanks!

jest.config.packages.js
ccharly: We used the same pattern than for other configuration fields in this file:
- https://github.com/MetaMask/accounts/pull/55/files/54eb2a99f2b8bcf0396ac1cbe21621a8a70643fd#diff-67d7eaa7b8718c9f7f2fb10f5be717922c0175ecd68f4af4c3520e505580eb4bR146-R147

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
ccharly: See:
- https://github.com/MetaMask/accounts/pull/55/commits/006878f9776c5ce4810e608175af8ab6883bc42b
- https://github.com/MetaMask/accounts/pull/55/commits/f916f75ce252c5336f995ad77e8d12e0eba56e77

packages/keyring-api/src/rpc-handler.ts
danroc: That's right, the type of the `request.method` field is `string` but the `case` values were `KeyringRpcMethod`, which caused the linter to complain.

I tought about changing the type of the `method` field, but then the linter would complain about the `default` case.

Another option I tried was to change this assertion:

```ts
assert(request, JsonRpcRequestStruct);
```

To something like (not the exact syntax):

```ts
assert(request, GetAccountRequestStruct | CreateAccountRequestStruct | ...);
```

But the type would not resolve correctly (I'm not sure if it's a Superstruct or TypeScript bug).

gantunesr: Interesting, I would assume that the type of an `enum` member to be a specific type (`string`, `number`, `boolean`, etc). It does seems to be confusing but in a minor scale, it can be improved in a future PR.

packages/keyring-api/src/superstruct.test-d.ts
```diff
@@ -4,6 +4,7 @@ import { expectAssignable, expectNotAssignable } from 'tsd';
 
 import { exactOptional, object } from '.';
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
gantunesr, line 7: Should we keep this unused vars?

packages/keyring-api/src/superstruct.ts
```diff
@@ -14,6 +14,7 @@ import type {
   Simplify,
 } from '@metamask/superstruct';
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
gantunesr, line 17: Same here regarding unused vars

packages/keyring-eth-ledger-bridge/test/document.shim.ts
```diff
@@ -1,24 +1,31 @@
...
 let documentShim: any;
 
+type Element = {
```
gantunesr, line 4: nit

```suggestion
type IframeElement = {
```

packages/keyring-eth-simple/src/simple-keyring.test.ts
```diff
@@ -21,8 +21,10 @@ import {
 import { add0x, Hex } from '@metamask/utils';
 import assert from 'assert';
 import { keccak256 } from 'ethereum-cryptography/keccak';
+// eslint-disable-next-line @typescript-eslint/naming-convention
 import OldEthereumTx from 'ethereumjs-tx';
 
+// eslint-disable-next-line @typescript-eslint/naming-convention
 import SimpleKeyring from '.';
```
gantunesr, line 28: Why is the import different here compared with the file `packages/keyring-eth-simple/src/sample.ts`?

```ts
import SimpleKeyring from './simple-keyring';
```

packages/keyring-eth-simple/src/simple-keyring.ts
```diff
@@ -27,8 +27,14 @@ type KeyringOpt = {
   version?: SignTypedDataVersion | string;
 };
 
+type Wallet = {
+  privateKey: Buffer;
+  publicKey: Buffer;
+};
```
gantunesr, line 33: nit suggestion,

```suggestion
type KeyPair = {
  privateKey: Buffer;
  publicKey: Buffer;
};
```

packages/keyring-eth-simple/src/simple-keyring.test.ts
```diff
@@ -21,8 +21,10 @@ import {
 import { add0x, Hex } from '@metamask/utils';
 import assert from 'assert';
 import { keccak256 } from 'ethereum-cryptography/keccak';
+// eslint-disable-next-line @typescript-eslint/naming-convention
```
gantunesr, line 24: What is the naming convention error here? Asking because I don't see a relevant difference with the imports from other files

gantunesr (commented):
Minor nits!

packages/keyring-api/src/superstruct.test-d.ts
ccharly: We need it to infer the type `ExactOptionalObject` and this file is mostly used for type testing, so we need to keep it.

packages/keyring-api/src/superstruct.ts
ccharly: Same remark here, it's only used "statically" (for the `type ExactOptionalTag`, so we need to keep it

packages/keyring-eth-ledger-bridge/test/document.shim.ts
ccharly: I must say I do prefer `Element` to match the `createElement` name here, is that ok for you?

packages/keyring-eth-simple/src/simple-keyring.ts
ccharly: Even though the name might be better, I think that we should keep the original name (this PR was meant to only fix eslint warning/errors, and this one is not). Is that ok?

packages/keyring-eth-simple/src/simple-keyring.test.ts
ccharly: I believe that's because "default" exports should be named using camelCase naming convention, but here it's using PascalCase instead.

You are allowed to use PascalCase if you use "destructuring import" (not sure what's the official name for this).

packages/keyring-eth-ledger-bridge/test/document.shim.ts
gantunesr: is just a nit, not going to die on this hill

packages/keyring-eth-simple/src/simple-keyring.ts
gantunesr: sure

packages/keyring-eth-simple/src/simple-keyring.test.ts
ccharly: I did change the one in `sample.ts` to match the original import path!

gantunesr (approved):

PR #60, opened 2024-09-25
feat: add qr keyring package

mikesposito:
This is a WIP implementation of a Keyring compatible with Keystone QR hardware


mikesposito:
@danroc this PR is not moving the package from another source - this is a new implementation, and only shares a similar account derivation strategy from URs with the [original keyring package](https://github.com/KeystoneHQ/keystone-sdk-base/blob/master/packages/metamask-airgapped-keyring/src/MetaMaskKeyring.ts) currently used on clients

Do you think we should bring in it's commit history anyway? Also, the package is in a Keystone monorepo, which would include commits to other packages potentially unrelated

mikesposito:
@metamaskbot publish-previews

mikesposito:
@metamaskbot publish-preview

mikesposito:
@metamaskbot publish-preview

mikesposito:
@metamaskbot publish-preview

packages/keyring-eth-qr/package.json
```diff
@@ -0,0 +1,97 @@
+{
+  "name": "@metamask/eth-qr-keyring",
+  "version": "1.0.0",
```
mcmire, line 3: Should this start off at 1.0.0? I'm not sure how releases work in this monorepo but if it's the same as `core` then we would set the initial version in a release PR.

``` suggestion
  "version": "0.0.0",

File: packages/keyring-eth-qr/src/device.ts [hunk 2]
```diff
@@ -0,0 +1,613 @@
...
+  constructor(requestScan: QrKeyringBridge['requestScan']) {
+    this.#requestScan = requestScan;
+  }
+
+  /**
+   * Initialize the Device with a source.
+   *
+   * @param source - The account source, which can be a CryptoAccount, CryptoHDKey, or a UR string.
+   */
+  init(source: DeviceDetails | string | SerializedUR): void {
```
mcmire, line 203: What if `init` is called more than once? Should we check to see if the Device has already been initialized and do nothing or throw an error in that case? (I don't know how we usually handle this type of thing elsewhere.)

Alternatively, does it make sense to move the initialization to the constructor?

mcmire (commented):
I tried to review this. I don't know that I can speak to the intricacies of the keyring class, but I will say that now that it's extracted from KeyringController it does seem fairly simple and it was a lot less complicated than I imagined it would be.

packages/keyring-eth-qr/jest.config.js
```diff
@@ -0,0 +1,29 @@
...
+  // The display name when running multiple projects
+  displayName,
+
+  // An array of regexp pattern strings used to skip coverage collection
+  coveragePathIgnorePatterns: ['./src/tests'],
+
+  // An object that configures minimum threshold enforcement for coverage results
+  coverageThreshold: {
+    global: {
+      branches: 85.71,
```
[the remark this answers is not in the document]
mikesposito: Definitely! I'm still working on it and aiming to 100% coverage, though I'm still missing some code branches

packages/keyring-eth-qr/jest.config.js
```diff
@@ -0,0 +1,29 @@
...
+const baseConfig = require('../../jest.config.packages');
+
+const displayName = path.basename(__dirname);
+
+module.exports = merge(baseConfig, {
+  // The display name when running multiple projects
+  displayName,
+
+  // An array of regexp pattern strings used to skip coverage collection
+  coveragePathIgnorePatterns: ['./src/tests'],
```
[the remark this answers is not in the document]
mikesposito: Good catch! I copied this from another package of this monorepo, I see that they all have this line, but none of them have the `src/tests` directory. I removed it in f88fa3943985.

packages/keyring-eth-qr/package.json
mikesposito: Definitely, done in 5ab4c618aa82

packages/keyring-eth-qr/src/device.ts
mcmire: Nice, looks a lot better!

mikesposito:
@metamaskbot publish-preview

mikesposito:
@metamaskbot publish-preview

mikesposito:
@metamaskbot publish-preview

packages/keyring-eth-qr/src/device.ts
```diff
@@ -341,15 +341,14 @@ export class Device {
         ? DataType.transaction
         : DataType.typedTransaction;
 
-    const messageToSign = Buffer.from(
-      transaction.type === TransactionType.Legacy
-        ? RLP.encode(transaction.getMessageToSign())
-        : (transaction as FeeMarketEIP1559Transaction).getMessageToSign(),
+    const messageToSign = transaction.getMessageToSign();
+    const rawTransaction = Buffer.from(
+      Array.isArray(messageToSign) ? RLP.encode(messageToSign) : messageToSign,
```
mikesposito, line 346: I initially kept this equal to how it is done in the original Keystone package, though we encode the transaction in the same way it is done in the Ledger keyring package.

packages/keyring-eth-qr/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
```
cryptodev-2s, line 8: Should we add a reference to this initial PR ?

cryptodev-2s (approved):
LGTM! I’ll just leave one small suggestion for the changelog so we won’t need to reference the initial PR later. Everything else looks good!

packages/keyring-eth-qr/CHANGELOG.md
mikesposito: Right, good catch - added in [5a3b4bc](https://github.com/MetaMask/accounts/pull/60/commits/5a3b4bced00074ec598d5541b74e9084a6123ce2)

cryptodev-2s (approved):
LGTM!

gantunesr (approved):
Looks good

PR #63, opened 2024-09-27
fix!(keyring-eth-ledger-bridge): change `addAccounts` to return new accounts only

mikesposito:
The `addAccounts` method should return newly create accounts, instead of all the accounts present in the keyring.

Related: https://github.com/MetaMask/accounts-planning/issues/615


packages/keyring-eth-ledger-bridge/src/ledger-keyring.test.ts
mikesposito, line 1: Tests were passing regardless of the issue so even if we are not adding any new test case,  they have been changed to also cover this behavior

danroc (approved):

PR #64, opened 2024-09-30
fix!(keyring-eth-trezor): change `addAccounts` to return new accounts only

mikesposito:
The `addAccounts` method should return newly create accounts, instead of all the accounts present in the keyring.

Related: https://github.com/MetaMask/accounts-planning/issues/615


packages/keyring-eth-trezor/src/trezor-keyring.test.ts
```diff
@@ -237,7 +237,10 @@ describe('TrezorKeyring', function () {
         keyring.setAccountToUnlock(0);
         await keyring.addAccounts();
         keyring.setAccountToUnlock(2);
-        const accounts = await keyring.addAccounts();
+
```
ccharly, line 240: Moving it closer to its `setAccountToUnlock`:

```suggestion
```

ccharly (approved):

PR #66, opened 2024-10-03
release: 4.0.0

mikesposito:
## Description

This is the release candidate for version 4.0.0. See the changelogs for more details.

### DIFF
--- a/packages/keyring-eth-ledger-bridge/CHANGELOG.md
+++ b/packages/keyring-eth-ledger-bridge/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [5.0.0]
+
 ### Uncategorized
 
 - fix!(keyring-eth-ledger-bridge): change `addAccounts` to return new accounts only ([#63](https://github.com/MetaMask/accounts/pull/63))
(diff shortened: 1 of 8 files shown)


packages/keyring-eth-ledger-bridge/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [5.0.0]
+
+### Changed
+
+- **BREAKING**: `addAccounts` returns newly created accounts only ([#63](https://github.com/MetaMask/accounts/pull/63))
```
ccharly, line 14:
```suggestion
- **BREAKING**: `addAccounts` will now only return newly created accounts ([#63](https://github.com/MetaMask/accounts/pull/63))
  - This keyring was initially returning every accounts (previous and new ones), which is different from what is expected in the [`Keyring` interface].(https://github.com/MetaMask/utils/blob/v9.2.1/src/keyring.ts#L65)
```

packages/keyring-eth-trezor/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [4.0.0]
+
+### Changed
+
+- **BREAKING**: `addAccounts` returns newly created accounts only ([#64](https://github.com/MetaMask/accounts/pull/64))
```
ccharly, line 14:
```suggestion
- **BREAKING**: `addAccounts` will now only return newly created accounts ([#64](https://github.com/MetaMask/accounts/pull/64))
  - This keyring was initially returning every accounts (previous and new ones), which is different from what is expected in the [`Keyring` interface].(https://github.com/MetaMask/utils/blob/v9.2.1/src/keyring.ts#L65)
```

ccharly (commented):
WDYT about adding a bit more context here?

ccharly (approved):

PR #70, opened 2024-10-11
feat(snap-keyring-bridge): sanitize redirect URL

ccharly:
Sanitize redirect URL when using the Snap keyring bridge.

### DIFF
--- a/packages/keyring-snap-bridge/src/util.ts
+++ b/packages/keyring-snap-bridge/src/util.ts
@@ -78,3 +78,15 @@ export function throwError(message: string): never {
 export function equalsIgnoreCase(a: string, b: string): boolean {
   return a.toLowerCase() === b.toLowerCase();
 }
+
+/**
+ * Sanitize a URL.
+ *
+ * @param url - The URL to sanitize.
+ * @returns The new sanitized redirect URL.
+ */
+export function sanitizeUrl(url: string): string {
+  // We use a `URL` object to detect any badly formatted URL and to normalize/sanitize them.
+  const saferUrl = new URL(url);
+  return saferUrl.toString();
+}
--- a/packages/keyring-snap-bridge/src/SnapKeyring.ts
+++ b/packages/keyring-snap-bridge/src/SnapKeyring.ts
@@ -575,11 +575,30 @@ export class SnapKeyring extends EventEmitter {
     redirect: { message?: string; url?: string },
     snapId: SnapId,
   ): Promise<void> {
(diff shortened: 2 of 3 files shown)


ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

packages/keyring-snap-bridge/package.json
```diff
@@ -51,9 +52,6 @@
     "typedoc": "^0.25.13",
     "typescript": "~5.4.5"
   },
-  "peerDependencies": {
```
danroc, line 54: Why this change? I think we still use some `keyring-api` types in our public API 🤔

packages/keyring-snap-bridge/src/util.ts
```diff
@@ -78,3 +78,16 @@ export function throwError(message: string): never {
...
 }
+
+/**
+ * Sanitize a URL.
+ *
+ * @param url - The URL to sanitize.
+ * @returns The new sanitized redirect URL.
+ */
+export function sanitizeUrl(url: string): string {
+  // For now, we only re-create the URL object which should take care of most of the sanitizing, like replacing
```
danroc, line 89: Why "most" sanitizing? Do you have other examples where it's not sanitized correctly?

packages/keyring-snap-bridge/package.json
ccharly: That was mainly to make the preview builds work (so I can use them in the extension afterward)

But yes, I'll definitely revert those changes before merging 👍

packages/keyring-snap-bridge/src/util.ts
ccharly: Good question, the wording is probably not the best here.  Maybe something like:
```ts
// We use a `URL` object to detect any badly formatted URL and to normalize/sanitize them.
```

WDYT?

ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

packages/keyring-snap-bridge/package.json
ccharly: I did revert that back now!

danroc (approved):

PR #82, opened 2024-11-04
feat: adds Solana account type to keyring API

zone-live:
We are adding the `SolAccountType` to support Solana account creation.

### DIFF
--- a/packages/keyring-api/src/api/account.ts
+++ b/packages/keyring-api/src/api/account.ts
@@ -21,12 +21,20 @@ export enum BtcAccountType {
 }
 
 /**
+ * Supported Solana account types.
+ */
+export enum SolAccountType {
+  Eoa = 'solana:system',
+}
+
+/**
  * Supported account types.
  */
 export type KeyringAccountType =
   | `${EthAccountType.Eoa}`
   | `${EthAccountType.Erc4337}`
-  | `${BtcAccountType.P2wpkh}`;
+  | `${BtcAccountType.P2wpkh}`
+  | `${SolAccountType.Eoa}`;
 
 /**
  * A struct which represents a Keyring account object. It is abstract enough to
@@ -48,6 +56,7 @@ export const KeyringAccountStruct = object({
     `${EthAccountType.Eoa}`,
     `${EthAccountType.Erc4337}`,
     `${BtcAccountType.P2wpkh}`,
+    `${SolAccountType.Eoa}`,
   ]),
 
   /**
--- a/packages/keyring-api/src/internal/types.ts
+++ b/packages/keyring-api/src/internal/types.ts
@@ -1,12 +1,13 @@
 import type { Infer, Struct } from '@metamask/superstruct';
 import { boolean, string, number } from '@metamask/superstruct';
 
-import { BtcAccountType, EthAccountType, KeyringAccountStruct } from '../api';
+import { BtcAccountType, EthAccountType, KeyringAccountStruct, SolAccountType } from '../api';
 import { BtcP2wpkhAccountStruct } from '../btc/types';
 import { EthEoaAccountStruct, EthErc4337AccountStruct } from '../eth/types';
 import { exactOptional, object } from '../superstruct';
+import { SolEoaAccountStruct } from '../sol/types';
 
-export type InternalAccountType = EthAccountType | BtcAccountType;
+export type InternalAccountType = EthAccountType | BtcAccountType | SolAccountType;
 
 export const InternalAccountMetadataStruct = object({
   metadata: object({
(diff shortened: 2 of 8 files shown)


aganglada (dismissed):

packages/keyring-api/src/sol/types.ts
```diff
@@ -0,0 +1,44 @@
...
+  'SolAddress',
+  /^[1-9A-HJ-NP-Za-km-z]{32,44}$/iu,
+);
+
+/**
+ * Supported Solana methods.
+ */
+export enum SolMethod {
+  // General transaction methods
+  SendSolana = 'sendSolana',
```
danroc, line 22: Is this method part of the standard Solana API? The `sendBitcoin` was added because there wasn't really a BTC standard method that we could use, but if there is one for Solana we should use it.

packages/keyring-api/src/sol/types.ts
```diff
@@ -0,0 +1,44 @@
+import type { Infer } from '@metamask/superstruct';
+import { array, enums, literal } from '@metamask/superstruct';
+
+import { KeyringAccountStruct, SolAccountType } from '../api';
+import { object, definePattern } from '../superstruct';
+
+/**
+ * Solana addresses are encoded using Base58
+ * Represented as 32 bytes in the format of an Ed25519 PublicKey
+ * And a valid Solana address is 44 characters long
```
danroc, line 10: It doesn't match the regex, did you mean it can be between 32 and 44 characters long?

zone-live: yes, they are normally 44 but can also be 32

zone-live: I'll edit the comment 👍🏼

packages/keyring-api/src/api/account.ts
```diff
@@ -20,13 +20,21 @@ export enum BtcAccountType {
   P2wpkh = 'bip122:p2wpkh',
 }
 
+/**
+ * Supported Solana account types.
+ */
+export enum SolAccountType {
+  Eoa = 'solana:system',
```
ccharly, line 27: Is this `system` a known thing in Solana? Since we are naming the value `Eoa`, I was wondering if this should rather be:

```ts
Eoa = 'solana:eoa',
```

Or:

```ts
System = 'solana:system',
```

packages/keyring-api/src/sol/types.ts
```diff
@@ -0,0 +1,44 @@
+import type { Infer } from '@metamask/superstruct';
+import { array, enums, literal } from '@metamask/superstruct';
+
+import { KeyringAccountStruct, SolAccountType } from '../api';
+import { object, definePattern } from '../superstruct';
+
+/**
+ * Solana addresses are encoded using Base58
+ * Represented as 32 bytes in the format of an Ed25519 PublicKey
+ * And a valid Solana address is usually 44 characters long but can also be 32
```
ccharly, line 10: Could we re-phrase it like, WDYT @danroc?

```suggestion
 * Solana addresses are represented in the format of a 256-bit ed25519 public key and
 * are encoded using base58.
 * They are usually 32 to 44 characters long.
```

packages/keyring-api/src/api/account.ts
zone-live: Hehehe I had it like that `solana:eoa` at the begining 😆 but changed to be `solana:system` since it's created by the System Program and referred as a "System Account" in some [docs](https://solana.com/docs/core/accounts) of solana. But for this case I also think that we can go with `Eoa = 'solana:eoa', It's clearer. 👍🏼 
`

packages/keyring-api/src/internal/types.ts
```diff
@@ -1,12 +1,21 @@
...
-import { BtcAccountType, EthAccountType, KeyringAccountStruct } from '../api';
+import {
+  BtcAccountType,
+  EthAccountType,
+  KeyringAccountStruct,
+  SolAccountType,
+} from '../api';
 import { BtcP2wpkhAccountStruct } from '../btc/types';
 import { EthEoaAccountStruct, EthErc4337AccountStruct } from '../eth/types';
+import { SolEoaAccountStruct } from '../sol/types';
```
ccharly, line 12: Should be renamed `SolAccountDataAccountStruct`, but let's wait for @danroc's input on the name first!

packages/keyring-api/src/sol/types.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Supported Solana methods.
+ */
+export enum SolMethod {
+  // General transaction methods
+  SendAndConfirmTransaction = 'sendAndConfirmTransaction',
+}
+
+export const SolEoaAccountStruct = object({
```
ccharly, line 25: Should be renamed `SolAccountDataAccountStruct` here too

packages/keyring-api/src/sol/types.ts
```diff
@@ -0,0 +1,44 @@
...
+   */
+  type: literal(`${SolAccountType.DataAccount}`),
+
+  /**
+   * Account supported methods.
+   */
+  methods: array(enums([`${SolMethod.SendAndConfirmTransaction}`])),
+});
+
+export type SolEoaAccount = Infer<typeof SolEoaAccountStruct>;
```
ccharly, line 44: Also here

danroc (dismissed):

ccharly (approved):

PR #84, opened 2024-11-06
Release/8.0.0

zone-live:
This is the release candidate for version 8.0.0. See the changelogs for more details, the goal is to release `keyring-api`.

### DIFF
--- a/packages/keyring-api/CHANGELOG.md
+++ b/packages/keyring-api/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [10.0.0]
+
 ### Uncategorized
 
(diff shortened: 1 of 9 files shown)


ulissesferreira (approved):

packages/keyring-api/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [10.0.0]
+
+### Uncategorized
```
danroc, line 12: We should have `Uncategorized` in the changelog (we need to check why the CI didn't fail).

packages/keyring-api/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [10.0.0]
```
danroc, line 10: There is no breaking change, we should bump only the minor version.

PR #86, opened 2024-11-06
fix(11215): device id should be cleared up when forgetting a device

Akaryatrh:
Currently when we call forgetDevice function from ledger keyring, it doesn't clear up the stored `deviceId`. This could be a blocker when checking for existing stored `deviceId` client side.

Then, this will unblock a [fix for multiple devices handling](https://github.com/MetaMask/metamask-mobile/pull/12127) as [an issue has been found during QA testing](https://github.com/MetaMask/metamask-mobile/issues/11215#issuecomment-2459557879) related to `deviceId`.


packages/keyring-eth-ledger-bridge/src/ledger-keyring.test.ts
```diff
@@ -580,6 +580,17 @@ describe('LedgerKeyring', function () {
         expect(keyring.isUnlocked()).toBe(false);
         expect(accounts).toHaveLength(0);
       });
+
+      it('deviceId should not be stored aymore', async function () {
```
ccharly, line 584:
```suggestion
      it('deviceId should not be cleared after forgetting the device', async function () {
```

ccharly (approved):

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -551,6 +551,7 @@ export class LedgerKeyring extends EventEmitter {
   }
 
   forgetDevice(): void {
+    this.deviceId = '';
```
mikesposito, line 554: should this be `null` instead?

ccharly: I've seen other mention to `''` when deserializing

Akaryatrh: Well, its initial state is an empty string, so I don't know if we want to change on both sides?

mikesposito (approved):

PR #90, opened 2024-11-08
fix: do not lower address for non-EVM accounts during account creation

ccharly:
Some non-EVM account's addresses might be case-sensitive like Solana.

We didn't have any issue with Bitcoin native segwit addresses (since they are lower-case already), but to avoid introducing this "lower-case" behavior everywhere we will only modify the address for EVM accounts.

cc: @aganglada

### DIFF
--- a/packages/keyring-snap-bridge/src/SnapKeyring.test.ts
+++ b/packages/keyring-snap-bridge/src/SnapKeyring.test.ts
@@ -104,10 +104,17 @@ describe('SnapKeyring', () => {
   };
   const btcP2wpkhAccount = {
     id: '11cffca0-12cc-4779-8f82-23273c062e29',
-    address: 'bc1qxy2kgdygjrsqtzq2n0yrf2493p83kkfjhx0wlh'.toLowerCase(),
+    address: 'bc1qxy2kgdygjrsqtzq2n0yrf2493p83kkfjhx0wlh',
     options: {},
     methods: [...Object.values(BtcMethod)],
     type: BtcAccountType.P2wpkh,
+  };
+  const solDataAccount = {
+    id: '780ee179-5ab5-449d-9c25-34e12c1ada66',
+    address: '3d4v35MRK57xM2Nte3E3rTQU1zyXGVrkXJ6FuEjVoKzM',
+    options: {},
+    methods: [...Object.values(SolMethod)],
+    type: SolAccountType.DataAccount,
   };
 
   const accounts = [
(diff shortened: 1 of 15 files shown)


ccharly:
@metamaskbot publish-previews

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -100,6 +101,21 @@ export type SnapKeyringCallbacks = {
...
+/**
+ * Normalize account's address.
+ *
+ * @param account - The account.
+ * @returns The normalized account's address.
+ */
+function normalizeAccountAddress(account: KeyringAccount): string {
+  // FIXME: Is it required to lowercase the address here? For now we'll keep this behavior
+  // only for Ethereum address and use the original one for other non-EVM accounts.
+  // For example, solana addresses are case-sensitives.
```
danroc, line 113:
```suggestion
  // For example, Solana addresses are case-sensitives.
```

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -100,6 +101,21 @@ export type SnapKeyringCallbacks = {
...
 
+/**
+ * Normalize account's address.
+ *
+ * @param account - The account.
+ * @returns The normalized account's address.
+ */
+function normalizeAccountAddress(account: KeyringAccount): string {
+  // FIXME: Is it required to lowercase the address here? For now we'll keep this behavior
+  // only for Ethereum address and use the original one for other non-EVM accounts.
```
danroc, line 112:
```suggestion
  // only for Ethereum addresses and use the original one for other non-EVM accounts.
```

gantunesr (approved):

zone-live (approved):

packages/keyring-snap-bridge/src/SnapKeyring.test.ts
```diff
@@ -592,9 +647,13 @@ describe('SnapKeyring', () => {
...
       const addresses = await keyring.getAccounts();
-      expect(addresses).toStrictEqual(
-        accounts.map((a) => a.address.toLowerCase()),
-      );
+      expect(addresses).toStrictEqual([
+        ethEoaAccount1.address.toLowerCase(),
+        ethEoaAccount2.address.toLowerCase(),
+        ethEoaAccount3.address.toLowerCase(),
+        ethErc4337Account.address.toLowerCase(),
+        btcP2wpkhAccount.address,
```
danroc, line 655: Can be handled in a different PR, but we should have a Solana account too.

PR #97, opened 2024-11-15
chore: bump typescript to version 5.6.3

ccharly:
Just upgrading to the latest stable TypeScript version.

This is also related to this comment: https://github.com/MetaMask/accounts/pull/40#issuecomment-2423429098.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "ts-jest": "^29.0.5",
     "ts-node": "^10.9.2",
     "typedoc": "^0.25.13",
-    "typescript": "~5.4.5"
+    "typescript": "~5.6.3"
   },
   "packageManager": "yarn@4.5.0",
   "engines": {
--- a/packages/keyring-api/package.json
+++ b/packages/keyring-api/package.json
@@ -71,7 +71,7 @@
     "ts-node": "^10.9.2",
     "tsd": "^0.31.0",
     "typedoc": "^0.25.13",
-    "typescript": "~5.4.5"
+    "typescript": "~5.6.3"
   },
   "peerDependencies": {
     "@metamask/providers": "^18.1.0"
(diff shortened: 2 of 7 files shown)


packages/keyring-eth-ledger-bridge/src/ledger-mobile-bridge.test.ts
```diff
@@ -25,7 +29,7 @@ describe('LedgerMobileBridge', function () {
   };
 
   const mockTransport = {
-    deviceModel: {} || null,
+    deviceModel: { id: '' } as MockDeviceModel, // Will be set by `beforeEach`
```
danroc, line 32:
```suggestion
    deviceModel: { id: '' } as MockDeviceModel, // `id` will be set by `beforeEach`
```

danroc (approved):

packages/keyring-eth-ledger-bridge/src/ledger-mobile-bridge.test.ts
```diff
@@ -220,8 +224,8 @@ describe('LedgerMobileBridge', function () {
       );
     });
 
-    it('throws error when transport.deviceMode is not set', async function () {
-      mockTransport.deviceModel = null;
+    it('throws error when transport.deviceModel is not set', async function () {
+      mockTransport.deviceModel = null as unknown as MockDeviceModel;
```
gantunesr, line 228: Instead of typecasting, do you think it's better to just add a directive to ignore the error?

```suggestion
    it('throws error when transport.deviceModel is not set', async function () {
      // // @ts-ignore deviceModel must be null to validate test case
      mockTransport.deviceModel = null;
```

ccharly: I think I prefer your suggestion indeed, lemme change that!

gantunesr (approved):

PR #100, opened 2024-11-22
refactor(keyring-eth-hd)!: move seed generation to deserialization

FrederikBolding:
This PR refactors the `eth-hd-keyring` constructor to remove the deserialization step directly in the constructor. This lets us turn `deserialize` into a proper async function and use an async version of `mnemonicToSeed` (which should be swapped out for something faster in the future). 

This is a breaking change and requires that all usage of `eth-hd-keyring` calls `deserialize` with the arguments previously passed into the constructor. This seems to already be the case in the `KeyringController`, but may deserve a more thorough look.

This PR also changes the function signature of `generateRandomMnemonic`, making this an async function as well.

Appreciate any feedback and/or sanity checks on this!

### DIFF
--- a/packages/keyring-eth-hd/src/index.js
+++ b/packages/keyring-eth-hd/src/index.js
@@ -16,8 +16,7 @@ import {
 } from '@metamask/eth-sig-util';
 import {
   generateMnemonic,
-  // eslint-disable-next-line n/no-sync
-  mnemonicToSeedSync,
+  mnemonicToSeed,
   validateMnemonic,
 } from '@metamask/scure-bip39';
 import { wordlist } from '@metamask/scure-bip39/dist/wordlists/english';
@@ -32,14 +31,13 @@ const type = 'HD Key Tree';
 
 class HdKeyring {
   /* PUBLIC METHODS */
-  constructor(opts = {}) {
+  constructor() {
     this.type = type;
     this._wallets = [];
-    this.deserialize(opts);
   }
 
-  generateRandomMnemonic() {
-    this._initFromMnemonic(generateMnemonic(wordlist));
+  async generateRandomMnemonic() {
+    await this._initFromMnemonic(generateMnemonic(wordlist));
   }
 
   _uint8ArrayToString(mnemonic) {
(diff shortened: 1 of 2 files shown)


danroc:
I haven't reviewed all the changes in detail, but the idea makes sense to me. I’d like to get @mikesposito’s opinion as well, as my main concern is the potential impact on the interface, other keyrings, and the KeyringController.

packages/keyring-eth-hd/src/index.js
```diff
@@ -116,14 +114,14 @@ class HdKeyring {
     this.hdPath = opts.hdPath || hdPathString;
 
     if (opts.mnemonic) {
-      this._initFromMnemonic(opts.mnemonic);
+      await this._initFromMnemonic(opts.mnemonic);
```
mikesposito, line 117: Despite the title of this PR (`Move seed generation to deserialization`), I understand that the seed generation was already done using the `deserialize` method, though now we are not automatically calling `deserialize` during Keyring construction. Is this correct to say?

FrederikBolding: Yeah, the title is meant to imply that directly calling the deserialization function is the only way to trigger seed generation now, except for `generateRandomMnemonic`

FrederikBolding:
@mikesposito We will have to update the KeyringController to await `keyring.generateRandomMnemonic` at the very least.

packages/keyring-eth-hd/src/index.js
```diff
@@ -32,14 +31,13 @@ const type = 'HD Key Tree';
 
 class HdKeyring {
   /* PUBLIC METHODS */
-  constructor(opts = {}) {
+  constructor() {
     this.type = type;
     this._wallets = [];
-    this.deserialize(opts);
```
ccharly, line 38: This is purely cosmetic, but WDYT of adding some `static` helpers, to avoid having to explicitly call the `deserialize` method (after calling `new`)?

Something like this:
```ts
const keyring = await HdKeyring.fromOptions(opts);
```

This makes it slightly more natural IMO and also makes sure that `deserialize` is being called underneath.

And the implementation would be as simple as:
```ts
static async fromOptions(opts = {}) {
  const keyring = new HdKeyring();
  await keyring.deserialize(opts);
  return keyring;
}
```

I think this way, we still keep somewhat a "sort of" `async` constructor?

ccharly: Yes, I do think it's mostly on the tests, so it's probably not really that useful.

So let's ignore this for now. 🙈

ccharly (approved):

PR #119, opened 2024-12-06
fix: depcheck

ccharly:
`depcheck` was not running properly. It has to be invoked with `yarn workspaces foreach`.


packages/keyring-eth-trezor/src/trezor-bridge.ts
```diff
@@ -11,6 +11,10 @@ import type {
   EthereumSignTypedHash,
 } from '@trezor/connect-web';
 
+// TrezorConnect.getPublicKey has two overloads
+// It is not possible to extract them from the library using utility types
+export type TrezorConnectPublicKey = { publicKey: string; chainCode: string };
```
danroc, line 16:
```suggestion
export type ExtendedPublicKey = { publicKey: string; chainCode: string };
```

danroc (approved):

PR #122, opened 2024-12-06
fix: missing references in some tsconfig.json

ccharly:
Some `tsconfig.json` are not aligned with their `tsconfig.build.json`. This might cause some dev tools (that rely on `tsconfig.json` only) to not work as expected.

### DIFF
--- a/packages/keyring-snap-client/tsconfig.json
+++ b/packages/keyring-snap-client/tsconfig.json
@@ -3,6 +3,14 @@
   "compilerOptions": {
     "baseUrl": "./"
   },
+  "references": [
+    {
+      "path": "../keyring-api/tsconfig.build.json"
+    },
+    {
+      "path": "../keyring-utils/tsconfig.build.json"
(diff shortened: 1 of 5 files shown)


packages/keyring-snap-client/tsconfig.json
```diff
@@ -3,6 +3,14 @@
   "compilerOptions": {
     "baseUrl": "./"
   },
+  "references": [
+    {
+      "path": "../keyring-api/tsconfig.build.json"
```
danroc, line 8: Did the linter check these files? I think the formatting is not consistent

ccharly: It should since we are using `lint:misc` in the main `lint` command. And running `yarn prettier packages/*/tsconfig*.json -w` does not change anything.

But I agree that it looks somehow inconsistent sometimes 😅

danroc (approved):

PR #131, opened 2024-12-10
refactor: move deps to devDeps for types + use peerDep for keyring-api

ccharly:
Updating some deps to avoid pulling non-needed dependencies in the main dependency tree.

Also, use `@metamask/keyring-api` as a `peerDependencies` for every packages to force using the same API everywhere.

### DIFF
--- a/packages/keyring-eth-hd/package.json
+++ b/packages/keyring-eth-hd/package.json
@@ -61,7 +61,8 @@
     "@ts-bridge/cli": "^0.6.1",
     "@types/jest": "^29.5.12",
     "deepmerge": "^4.2.2",
-    "jest": "^29.5.0"
+    "jest": "^29.5.0",
+    "typescript": "~5.6.3"
   },
   "engines": {
     "node": "^18.18 || >=20"
--- a/packages/keyring-eth-trezor/package.json
+++ b/packages/keyring-eth-trezor/package.json
@@ -52,8 +52,7 @@
     "@metamask/eth-sig-util": "^8.1.2",
     "@trezor/connect-plugin-ethereum": "^9.0.3",
     "@trezor/connect-web": "^9.1.11",
-    "hdkey": "^2.1.0",
-    "tslib": "^2.6.2"
+    "hdkey": "^2.1.0"
   },
   "devDependencies": {
     "@ethereumjs/common": "^3.2.0",
@@ -76,8 +75,12 @@
     "sinon": "^19.0.2",
     "ts-jest": "^29.0.5",
     "ts-node": "^10.9.2",
+    "tslib": "^2.6.2",
     "typedoc": "^0.25.13",
     "typescript": "~5.6.3"
+  },
+  "peerDependencies": {
+    "tslib": "^2.6.2"
   },
   "engines": {
     "node": "^18.18 || >=20"
--- a/packages/keyring-internal-api/package.json
+++ b/packages/keyring-internal-api/package.json
@@ -45,7 +45,6 @@
     "test:watch": "jest --watch"
   },
   "dependencies": {
-    "@metamask/keyring-api": "workspace:^",
     "@metamask/keyring-utils": "workspace:^",
     "@metamask/superstruct": "^3.1.0",
     "@metamask/utils": "^11.0.1"
@@ -54,6 +53,7 @@
     "@lavamoat/allow-scripts": "^3.2.1",
     "@lavamoat/preinstall-always-fail": "^2.1.0",
     "@metamask/auto-changelog": "^3.4.4",
+    "@metamask/keyring-api": "workspace:^",
     "@ts-bridge/cli": "^0.6.1",
     "@types/jest": "^29.5.12",
     "@types/node": "^20.12.12",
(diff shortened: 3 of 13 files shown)


packages/keyring-eth-trezor/package.json
```diff
@@ -76,9 +75,13 @@
     "sinon": "^19.0.2",
     "ts-jest": "^29.0.5",
     "ts-node": "^10.9.2",
+    "tslib": "^2.6.2",
     "typedoc": "^0.25.13",
     "typescript": "~5.6.3"
   },
+  "peerDependencies": {
+    "tslib": "^2.6.2"
```
ccharly, line 83: This should have been a `peerDependencies` since the introduction of `@trezor/connect-web`.

ccharly:
@metamaskbot publish-preview

packages/keyring-eth-hd/package.json
```diff
@@ -61,7 +61,8 @@
     "@ts-bridge/cli": "^0.6.1",
     "@types/jest": "^29.5.12",
     "deepmerge": "^4.2.2",
-    "jest": "^29.5.0"
+    "jest": "^29.5.0",
+    "typescript": "~5.6.3"
```
ccharly, line 65: Not needed as of now, but since this package has already been "future-proof" for `ts-bridge` we might wanna add `typescript` as well.

packages/keyring-internal-api/package.json
```diff
@@ -68,6 +68,9 @@
     "typedoc": "^0.25.13",
     "typescript": "~5.6.3"
   },
+  "peerDependencies": {
+    "@metamask/keyring-api": "workspace:^"
```
ccharly, line 72: To enforce the relationship with the `keyring-api` we make it now a peer dep, so the consumer has to provide the correct `keyring-api` version on its own.

packages/keyring-internal-snap-client/package.json
```diff
@@ -45,19 +45,18 @@
     "test:watch": "jest --watch"
   },
   "dependencies": {
-    "@metamask/keyring-api": "workspace:^",
     "@metamask/keyring-snap-client": "workspace:^",
-    "@metamask/keyring-utils": "workspace:^",
-    "@metamask/snaps-controllers": "^9.10.0",
```
ccharly, line 51: We actually only need types from `snaps-*` packages, so we moved everything into `devDependencies`.

packages/keyring-internal-snap-client/package.json
```diff
@@ -71,10 +70,8 @@
     "ts-node": "^10.9.2",
     "tsd": "^0.31.0",
     "typedoc": "^0.25.13",
-    "typescript": "~5.6.3"
-  },
-  "peerDependencies": {
-    "@metamask/providers": "^18.3.1"
```
ccharly, line 77: No longer required, since `snaps-*` packages are node `devDependencies`.

packages/keyring-snap-bridge/package.json
```diff
@@ -39,27 +39,23 @@
...
-    "webextension-polyfill": "^0.12.0"
+    "uuid": "^9.0.1"
   },
   "devDependencies": {
-    "@lavamoat/allow-scripts": "^3.2.1",
-    "@lavamoat/preinstall-always-fail": "^2.1.0",
-    "@metamask/auto-changelog": "^3.4.4",
+    "@metamask/keyring-api": "workspace:^",
     "@metamask/providers": "^18.3.1",
+    "@metamask/snaps-controllers": "^9.10.0",
```
ccharly, line 52: Same here, we only types from `snaps-*`.

packages/keyring-snap-bridge/package.json
```diff
@@ -68,11 +64,11 @@
...
     "tsd": "^0.31.0",
     "typedoc": "^0.25.13",
-    "typescript": "~5.6.3"
+    "typescript": "~5.6.3",
+    "webextension-polyfill": "^0.12.0"
   },
   "peerDependencies": {
-    "@metamask/keyring-api": "workspace:^",
-    "@metamask/providers": "^18.3.1"
+    "@metamask/keyring-api": "workspace:^"
```
ccharly, line 71: Same remark here.

packages/keyring-eth-trezor/package.json
ccharly: I tried to follow the same pattern than for `core` which is to re-declare (explicitly) `peerDependencies` in the package if we are not providing it.

packages/keyring-internal-snap-client/package.json
```diff
@@ -45,18 +45,17 @@
...
     "@metamask/snaps-sdk": "^6.7.0",
-    "@metamask/snaps-utils": "^8.3.0",
-    "webextension-polyfill": "^0.12.0"
+    "@metamask/snaps-utils": "^8.3.0"
   },
   "devDependencies": {
     "@lavamoat/allow-scripts": "^3.2.1",
     "@lavamoat/preinstall-always-fail": "^2.1.0",
     "@metamask/auto-changelog": "^3.4.4",
+    "@metamask/keyring-api": "workspace:^",
```
ccharly, line 58: Moved to `dev` since we only need the `keyring-api` for tests here.

PR #147, opened 2025-01-16
feat(keyring-snap-bridge): add Snap keyring migrations

ccharly:
With the new additions of the `scopes` field on the `KeyringAccount`, the Snap keyring state might be incompatible when we first deserialize it (and until it's being updated).

We initially wanted to run those migrations on the client side (extension) to update the `AccountsController` internal state, but the Snap keyring state will still be outdated. Also, the `AccountsController` is re-creating its internal state based on the accounts coming from each keyrings (including Snap accounts), so the Snap keyring state HAS TO BE updated, since it's the "source of truth" here.

### DIFF
--- a/packages/keyring-snap-bridge/src/SnapKeyring.test.ts
+++ b/packages/keyring-snap-bridge/src/SnapKeyring.test.ts
@@ -16,6 +16,8 @@ import {
   EthMethod,
   SolMethod,
   KeyringEvent,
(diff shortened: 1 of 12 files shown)


ccharly:
@metamaskbot publish-preview

darioAnongba (dismissed):

danroc (approved):

PR #148, opened 2025-01-16
feat: add `listAccountAssets` method

ccharly:
This new method allows a Snap account to list all the owned assets for a given accounts.

It can be used in combinations with `getAccountBalances`.

The client would first fetch the assets for an account and then get the associated balances:
```ts
// Keyring client to interact with the account Snap.
const client = new KeyringClient(...);

// The account we want to retrieve assets from.
const id = '418b6f26-40e7-4b50-a768-018c9aba496d';

// Assets.
const tokens: CaipAssetType[] = [];
const nfts: CaipAssetId[] = [];

// Fetch assets:
for (const asset in await client.listAccountAssets()) {
  if (isCaipAssetId(asset)) {
    nfts.push(asset);
  } else if (isCaipAssetType(asset)) {
    tokens.push(asset);
  } else {
    console.warn(`Malformed asset, skipping: ${asset}`);
  }
}

// Fetch token balances:
const balances = await client.getAccountBalances(tokens);
...
```

### DIFF
--- a/packages/keyring-api/src/api/caip.ts
+++ b/packages/keyring-api/src/api/caip.ts
@@ -12,7 +12,7 @@ const CAIP_ASSET_TYPE_OR_ID_REGEX =
   /^(?<chainId>(?<namespace>[-a-z0-9]{3,8}):(?<reference>[-_a-zA-Z0-9]{1,32}))\/(?<assetNamespace>[-a-z0-9]{3,8}):(?<assetReference>[-.%a-zA-Z0-9]{1,128})(\/(?<tokenId>[-.%a-zA-Z0-9]{1,78}))?$/u;
 
 /**
- * A CAIP-19 asset type or asset ID identifier, i.e., a human-readable type of asset identifier.
+ * A CAIP-19 asset type identifier, i.e., a human-readable type of asset identifier.
  */
 export const CaipAssetTypeStruct = definePattern<CaipAssetType>(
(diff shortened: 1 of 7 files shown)


packages/keyring-api/src/api/caip.ts
```diff
@@ -1,19 +1,25 @@
...
-  CAIP_ASSET_ID_REGEX,
-  CAIP_ASSET_TYPE_REGEX,
 } from '@metamask/utils';
 
+const CAIP_ASSET_TYPE_OR_ID_REGEX =
+  /^(?<chainId>(?<namespace>[-a-z0-9]{3,8}):(?<reference>[-_a-zA-Z0-9]{1,32}))\/(?<assetNamespace>[-a-z0-9]{3,8}):(?<assetReference>[-.%a-zA-Z0-9]{1,128})(\/(?<tokenId>[-.%a-zA-Z0-9]{1,78}))?$/u;
+
 /**
- * A CAIP-19 asset type identifier, i.e., a human-readable type of asset identifier.
+ * A CAIP-19 asset type or asset ID identifier, i.e., a human-readable type of asset identifier.
```
danroc, line 15: Is this comment correct? It looks like it's outdated.

ccharly: It's not 🙈 Good catch!

danroc (approved):

PR #151, opened 2025-01-17
fix: `init` uses a timeout and other methods reattempt `init`

mikesposito:
The `LedgerBridgeIframe` class gets initialized through the `init` method. However, if the iframe that is set up during this method call fails for some reason, the promise returned by `init` will never resolve/reject. 

Downstream, this prevents the KeyringController mutex from being ever released, causing deadlocks at unlock time.

The solution applied in this PR is to make the iframe set up during the `init` call optional, and then later reattempting it whenever another method needs the iframe and finds it uninitialized

### DIFF
--- a/packages/keyring-eth-ledger-bridge/src/ledger-iframe-bridge.ts
+++ b/packages/keyring-eth-ledger-bridge/src/ledger-iframe-bridge.ts
@@ -341,32 +341,23 @@ export class LedgerIframeBridge
   }
 
   async #setupIframe(bridgeUrl: string): Promise<void> {
-    const timeout = new Promise((_, reject) => {
-      setTimeout(() => {
+    return new Promise((resolve, reject) => {
+      const timeout = setTimeout(() => {
         reject(new Error('Bridge initialization timed out'));
       }, IFRAME_INIT_TIMEOUT);
-    });
 
-    const initIframe = new Promise<void>((resolve, reject) => {
       this.iframe = document.createElement('iframe');
       this.iframe.src = bridgeUrl;
       this.iframe.allow = `hid 'src'`;
+
       this.iframe.onload = async (): Promise<void> => {
+        clearTimeout(timeout);
         this.iframeLoaded = true;
         resolve();
       };
-      this.iframe.onerror = (): void => {
-        if (this.iframe) {
-          document.head.removeChild(this.iframe);
-          this.iframe = undefined;
-        }
-        this.iframeLoaded = false;
-        reject(new Error('Failed to load iframe'));
-      };
+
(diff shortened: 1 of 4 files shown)


packages/keyring-eth-ledger-bridge/src/ledger-iframe-bridge.ts
```diff
@@ -277,17 +298,38 @@ export class LedgerIframeBridge
...
         this.iframeLoaded = true;
         resolve();
       };
+      this.iframe.onerror = (): void => {
+        if (this.iframe) {
+          document.head.removeChild(this.iframe);
+        }
+        this.iframeLoaded = false;
+        reject(new Error('Failed to load iframe'));
+      };
```
mikesposito, line 328: I'm not sure that this is actually needed, since the timeout would kick in anyway, but seemed useful

packages/keyring-eth-ledger-bridge/src/ledger-iframe-bridge.ts
```diff
@@ -169,13 +186,12 @@ export class LedgerIframeBridge
...
   async updateTransportMethod(transportType: string): Promise<boolean> {
-    return new Promise((resolve, reject) => {
-      // If the iframe isn't loaded yet, let's store the desired transportType value and
-      // optimistically return a successful promise
-      if (!this.iframeLoaded) {
-        throw new Error('The iframe is not loaded yet');
-      }
+    // If the iframe isn't loaded yet, the initialization is reattempted
+    if (!this.iframeLoaded) {
+      await this.#init();
```
Gudahtt, line 191: Hmm, we certainly wouldn't want to trigger the creation of the iframe multiple times. Perhaps we could store a pending initialization as a promise somewhere, to prevent concurrent initialization attempts?

mikesposito: Good point. I added a deferred promise in https://github.com/MetaMask/accounts/pull/151/commits/fb00879dd8ffdadac8b524515c799a31ab449dba

mikesposito:
@metamaskbot publish-preview

mikesposito:
After some testing with preview builds, it appears that `onload` is not a reliable source to resolve the iframe initialization, as it is also called when the iframe load is errored. The changes in this PR are still valid against scenarios of slow connection, as the timeout would kick in before `onload`, but not if the page responds with an error. Additionally, after the iframe is loaded with an error, the rest of the methods will try to send messages to the iframe that will never be resolved, creating other scenarios of infinite loading.

A solution could be having a message handler on the bridge to check if the connection is alive (i.e. the iframe is initialized for real)

PR #152, opened 2025-01-20
feat!: use `Messenger` instead of `SnapsController`

ccharly:
Introducing the concept of `Messenger` (messaging system that we use in our `core` controllers) to break the runtime dependency we have with the `SnapController`.

We will also use this new `Messenger` later when the Snap keyring will re-forward some Snap account events to some other controllers.

This is **BREAKING** because of the removal of the `KeyringSnapControllerClient` and because the `SnapKeyring`'s constructor now requires a `Messenger` object instead of the `SnapController`.


packages/keyring-internal-snap-client/src/KeyringInternalSnapClient.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      expect(messenger.call).toHaveBeenCalledWith(
+        'SnapController:handleRequest',
+        request,
+      );
+      expect(accounts).toStrictEqual(accountsList);
+    });
+
+    it('calls the listAccounts method and return the result (withSnapId)', async () => {
+      const client = new KeyringInternalSnapClient({
+        messenger: messenger as unknown as KeyringInternalSnapClientMessenger,
```
danroc, line 56: Is this double-cast still needed?

danroc (approved):

PR #154, opened 2025-01-21
feat: add `account{AssetList,Balances,Transactions}Updated` keyring events + re-publish

ccharly:
Now that the `SnapKeyring` uses a `Messenger` to interact with some controllers (`SnapController` only for now). We can re-use this messenger to forward new events coming out from the Snap.

Those events will be mainly consumed by the new `Multichain*Controller`(s). This allows to use a "pub/sub pattern" rather than relying on polling information out from the Snap.

All those information can still be fetched explicitly from the Snap:
- `accountBalancesUpdated` -> `Keyring.getAccountBalances`
- `accountTransactionsUpdated` -> `Keyring.listAccountTransactions`
- `accountAssetListUpdated` -> `OnAssetLookup` (through [SIP-29](https://github.com/MetaMask/SIPs/pull/154), once available)


packages/keyring-snap-bridge/src/SnapKeyring.test.ts
```diff
@@ -225,8 +186,10 @@ describe('SnapKeyring', () => {
 
   // Fake the ControllerMessenger and registers all mock actions here:
   const controllerMessenger: ControllerMessenger<
-    SnapKeyringAllowedActions,
-    SnapKeyringAllowedEvents
+    // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
+    SnapKeyringActions | AllowedActions,
+    // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
+    SnapKeyringEvents | AllowedEvents
```
ccharly, line 192: Eslint needs to be disabled here since `SnapKeyringActions` and `AllowedEvents` are both set to `never` meaning they'll get overridden by the other union's members.

I wanted to keep the union syntax here, since that's what we use in every other controllers.

packages/keyring-api/src/api/account.ts
```diff
@@ -45,6 +45,11 @@ export type KeyringAccountType =
   | `${BtcAccountType.P2wpkh}`
   | `${SolAccountType.DataAccount}`;
 
+/**
+ * Account ID (UUIDv4).
+ */
+export const AccountIdStruct = UuidStruct; // Alias for better naming purposes.
```
danroc, line 51: I'm ok with this definition being here, but we could consider having it in `keyring-utils` (I have no preference, just raising the point)

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -349,6 +352,48 @@ export class SnapKeyring extends EventEmitter {
...
+   *
+   * @param message - Event message.
+   * @returns `null`.
+   */
+  async #handleAccountTransactionsUpdated(message: SnapMessage): Promise<null> {
+    assert(message, AccountTransactionsUpdatedEventStruct);
+    const event = message.params;
+
+    this.#messenger.publish('SnapKeyring:accountTransactionsUpdated', event);
+    return null;
```
danroc, line 394: I wonder if it's worth to refactor these handlers to reuse the same code since only the struct and event name are different. WDYT?

packages/keyring-snap-bridge/src/SnapKeyring.test.ts
danroc: I'm on the fence here. I think we should just remove the `never` types and add an union in the future if needed. WDYT?

packages/keyring-snap-bridge/src/SnapKeyring.ts
ccharly: Yup, I kinda thought about the same TBH 😄 I'll do that!

packages/keyring-snap-bridge/src/SnapKeyringMessenger.ts
```diff
@@ -1,14 +1,48 @@
...
 
 export type SnapKeyringMessenger = RestrictedControllerMessenger<
-  'SnapKeyringMessenger',
-  SnapKeyringAllowedActions,
-  SnapKeyringAllowedEvents,
-  SnapKeyringAllowedActions['type'],
-  SnapKeyringAllowedEvents['type']
+  'SnapKeyring',
+  // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
+  AllowedActions | SnapKeyringActions,
```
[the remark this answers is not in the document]
ccharly: Sounds good, I can do that. I just wanted them to be there since that's the pattern we use everywhere for controllers (but they do have actions and events defined 🙈)

I wanted to "future-proof" the type declaration, but indeed that requires those eslint directives (which is bad).

I'll update that too!

packages/keyring-api/src/api/account.ts
ccharly: At least it would be closer to `UuidStruct` so I'm ok with that.

I just thought that `AccountId` types being part of the `keyring-api` made sense too!

danroc (dismissed):

ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

danroc (approved):

PR #155, opened 2025-01-21
release: 21.0.0

ccharly:
# Description

This is the release candidate for version 21.0.0. See the CHANGELOGs for more details.


packages/keyring-snap-bridge/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [8.1.1]
```
ccharly, line 10: This is only a patch bump, since the breaking changes to the `keyring-api` only affects keyring methods (which are not used by the bridge directly)

danroc (approved):

PR #157, opened 2025-01-22
feat: add `resolveAccountAddress` method + expose `SnapKeyring.submitRequest`

ccharly:
The new [SIP-26](https://github.com/MetaMask/SIPs/blob/main/SIPS/sip-26.md) requires some changes on the `SnapKeyring` and requires an additional keyring method to resolve account's address from a signing request.


packages/keyring-api/src/api/caip.test.ts
```diff
@@ -1,4 +1,40 @@
...
+    [],
+    '',
+    '!@#$%^&*()',
+    'foo',
+    'eip155',
+    'eip155:',
+  ])('returns false for an invalid chain id %s', (id) => {
+    expect(isCaipChainId(id)).toBe(false);
+  });
+});
```
ccharly, line 37: I re-exported this because of code coverage. I was not able to "skip" the re-exported types from it.

But we definitely needs to remove those tests now that we are using `@metamask/utils`!

packages/keyring-api/src/api/caip.ts
```diff
@@ -1,16 +1,28 @@
...
   isCaipAssetId,
 } from '@metamask/utils';
 
 const CAIP_ASSET_TYPE_OR_ID_REGEX =
   /^(?<chainId>(?<namespace>[-a-z0-9]{3,8}):(?<reference>[-_a-zA-Z0-9]{1,32}))\/(?<assetNamespace>[-a-z0-9]{3,8}):(?<assetReference>[-.%a-zA-Z0-9]{1,128})(\/(?<tokenId>[-.%a-zA-Z0-9]{1,78}))?$/u;
 
+/**
+ * A CAIP-2 chain ID, i.e., a human-readable agnostic chain ID.
+ */
+export const CaipChainIdStruct = definePattern<CaipChainId>(
```
FrederikBolding, line 19: This already exists in `utils`, do we need to redefine it?

ccharly: Not really, the only reason why we re-define those for now is because our `definePattern` gives better error messages, like here: https://github.com/MetaMask/accounts/blob/v23.0.0/packages/keyring-snap-sdk/src/rpc-handler.test.ts#L604

IMO, that would be great that we move those to `utils` at some point.

But type-wise, they truly are the same, yes.

packages/keyring-api/src/api/caip.test.ts
ccharly: See:
- https://github.com/MetaMask/utils/pull/228

packages/keyring-api/src/api/caip.ts
ccharly: See:
- https://github.com/MetaMask/utils/pull/228

packages/keyring-api/src/api/address.ts
```diff
@@ -0,0 +1,19 @@
+import type { Infer } from '@metamask/superstruct';
+import { object, string } from '@metamask/superstruct';
+
+/**
+ * The resolved address of an account.
```
danroc, line 5: I think it's worth mentioning that the address is resolved from a request and that it will be used by the RPC Router. Otherwise, it can be a bit confusing to understand what this type is used for.

packages/keyring-api/src/api/caip.test.ts
ccharly: Now available on `@metamask/utils@11.10`

packages/keyring-api/src/api/address.ts
ccharly: https://github.com/MetaMask/accounts/pull/157/commits/9c4ca153fb200535ab017bea55783338ee565016

packages/keyring-api/src/rpc.ts
```diff
@@ -178,6 +185,32 @@ export type GetAccountBalancesResponse = Infer<
...
 >;
 
+// ----------------------------------------------------------------------------
+// Resolve account address
+
+export const ResolveAccountAddressRequestStruct = object({
+  ...CommonHeader,
+  method: literal('keyring_resolveAccountAddress'),
+  params: object({
+    scope: CaipChainIdStruct,
```
danroc, line 195: Do we have the scope and is it needed? I wonder if we will have it for off-chain signing requests.

danroc: From [CAIP-27](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-27.md), we can assume that it's always available:

> **scope** (required) - a valid [CAIP-2](https://chainagnostic.org/CAIPs/caip-2) network identifier, previously authorized by or within a `scopeObject` in the active session

packages/keyring-api/src/api/keyring.ts
```diff
@@ -109,6 +111,23 @@ export type Keyring = {
...
+   * incoming non-EVM dapp signing requests.
+   *
+   * @param scope - Request's scope (CAIP-2).
+   * @param request - Signing request object.
+   * @returns A Promise that resolves to the account address that must
+   * be used to process this signing request, or null if none candidates
+   * could be found.
+   */
+  resolveAccountAddress(
+    id: string,
```
danroc, line 127:
```suggestion
    scope: CaipChainId,
```

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -534,58 +554,126 @@ export class SnapKeyring extends EventEmitter {
...
+   */
+  async resolveAccountAddress(
+    snapId: SnapId,
+    scope: CaipChainId,
+    request: JsonRpcRequest,
+  ): Promise<ResolvedAccountAddress | null> {
+    // We do check that the incoming Snap ID is known by the keyring.
+    if (!this.hasSnapId(snapId)) {
+      throw new Error(
+        `Unable to resolve account's address: unknown Snap ID: ${snapId}`,
```
danroc, line 584:
```suggestion
        `Unable to resolve account address: unknown Snap ID: ${snapId}`,
```

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -534,58 +554,126 @@ export class SnapKeyring extends EventEmitter {
     );
   }
 
+  /**
+   * Checks if a Snap ID is known from the keyring.
+   *
+   * @param snapId - Snap ID.
+   * @returns `true` if the Snap ID is known, `false` otherwise.
+   */
+  hasSnapId(snapId: SnapId): boolean {
```
danroc, line 563: Do you think this method will be used by others or can we make it private?

packages/keyring-api/src/api/keyring.ts
ccharly: Good catch 🙈 thanks!

packages/keyring-snap-bridge/src/SnapKeyring.ts
ccharly: My idea was to expose it so that we would have a way from a the `SnapKeyring` to know which Snaps are being used.

But that's just me maybe 🙈 We can make it private for now if you prefer!

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -534,58 +554,126 @@ export class SnapKeyring extends EventEmitter {
...
+    const { requestId, response } = await this.#submitSnapRequest({
+      snapId,
+      account,
+      method: method as AccountMethod,
+      params,
+      scope,
+    });
+
+    // For now, we enforce responses to be synchronous.
+    if (response.pending) {
```
danroc, line 627: I don't think we cannot enforce _all_ requests to be synchronous. We could add an optional argument as we did before.

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -534,58 +554,126 @@ export class SnapKeyring extends EventEmitter {
...
+   * This request cannot be an asynchronous keyring request.
+   *
+   * @param opts - Request options.
+   * @param opts.id - Account ID.
+   * @param opts.method - Method to call.
+   * @param opts.params - Method parameters.
+   * @param opts.scope - Selected chain ID (CAIP-2).
+   * @returns Promise that resolves to the result of the method call.
+   */
+  async submitRequest({
```
danroc, line 605: Is there a particular reason to have two implementations (`#submitSnapRequest` and `submitRequest`)?

ccharly: As discussed together, the main reasons were:
- SIP-26 uses an account ID + I wanted to prevent "async keyring request" in this context
- our internal use (private version of it) uses an account address instead

Also, the `#submitSnapRequest` is more like the "internal" version of it, that requires a `snapId` (that both `submitRequest` do no have as parameters).

packages/keyring-snap-bridge/src/SnapKeyring.ts
ccharly: As discussed internally, we will force this behavior for non-EVM requests (which are gonna be the sole consumer of this new method for now, through the multichain API + SIP-26)

ccharly:
@metamaskbot publish-preview

danroc (dismissed):

danroc (approved):

PR #159, opened 2025-01-24
fix!: make `scopes` more restrictive

ccharly:
Adding a more restricted `scopes` field to each account type definitions (`Struct`).

### DIFF
--- a/packages/keyring-api/src/btc/types.test.ts
+++ b/packages/keyring-api/src/btc/types.test.ts
@@ -1,4 +1,20 @@
-import { BtcP2wpkhAddressStruct } from './types';
+import { BtcScopes } from './constants';
+import type { BtcP2wpkhAccount } from './types';
+import {
+  BtcMethod,
+  BtcP2wpkhAccountStruct,
+  BtcP2wpkhAddressStruct,
+} from './types';
+import { BtcAccountType } from '../api';
+
+const MOCK_ACCOUNT = {
+  id: '55583f38-d81b-48f8-8494-fc543c2b5c95',
+  type: BtcAccountType.P2wpkh,
+  address: 'bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4',
+  methods: [BtcMethod.SendBitcoin],
+  options: {},
+  scopes: [BtcScopes.Mainnet],
+};
 
 describe('types', () => {
   describe('BtcP2wpkhAddressStruct', () => {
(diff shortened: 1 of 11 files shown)


ccharly:
@metamaskbot publish-preview

packages/keyring-snap-bridge/src/SnapKeyring.test.ts
```diff
@@ -117,7 +117,7 @@ describe('SnapKeyring', () => {
   };
   const ethEoaAccount3 = {
     id: 'c6697bcf-5710-4751-a1cb-340e4b50617a',
-    address: '0xab1G3q98V7C67T9103g30C0417610237A137d763'.toLowerCase(),
```
ccharly, line 120: `q` is not a valid character for an Ethereum address

packages/keyring-snap-bridge/src/SnapKeyring.test.ts
```diff
@@ -656,7 +656,7 @@ describe('SnapKeyring', () => {
           `Method '${EthMethod.PersonalSign}' not supported for account ${ethEoaAccount1.address}`,
         );
         // Restore `EthMethod.PersonalSign` and remove `EthMethod.SignTransaction`
-        updatedMethods = Object.values(EthMethod).filter(
```
ccharly, line 659: `EthMethod` was not truly correct here, since we use an EOA account here, which should use `ETH_EOA_METHODS` (since `EthMethod` also contains method for ERC4337 accounts)

packages/keyring-api/src/btc/types.ts
```diff
@@ -41,6 +49,11 @@ export const BtcP2wpkhAccountStruct = object({
    */
   type: literal(`${BtcAccountType.P2wpkh}`),
 
+  /**
+   * Account supported scopes (CAIP-2 chain IDs).
+   */
+  scopes: nonempty(array(CaipChainIdStruct)), // TODO: Would be great to further restrict this to be an array of 1 element?
```
danroc, line 55: I think it makes sense to add the restriction mentioned in the TODO. Any reason to not do it?

packages/keyring-snap-bridge/src/account.test.ts
```diff
@@ -0,0 +1,18 @@
+import type { KeyringAccount } from '@metamask/keyring-api';
+
+import { transformAccount } from './account';
+
+describe('account', () => {
+  it('throw for unknown account type', () => {
```
danroc, line 6:
```suggestion
  it('throws for unknown account type', () => {
```

packages/keyring-api/src/btc/types.ts
ccharly: No actual reason no, I'll check if `superstruct` has something for this, otherwise I might to write this logic myself!

ccharly: Actually, `size` should work, I just missed it when looking at `superstruct` lemme change that now!

packages/keyring-snap-bridge/src/migrations/v1.ts
```diff
@@ -40,10 +44,11 @@ export function getScopesForAccountV1(accountV1: KeyringAccountV1): string[] {
...
     case EthAccountType.Erc4337: {
       // EVM Erc4337 account
-      // NOTE: A Smart Contract account might not be compatible with every chain, but we still use
-      // "generic" scope for now. Also, there's no official Snap as of today that uses this account type. So
+      // NOTE: A Smart Contract account might not be compatible with every chain, in this case we just default
+      // to mainnet since we cannot really "guess" it from here.
+      // Also, there's no official Snap as of today that uses this account type. So
       // this case should never happen.
-      return [EthScopes.Namespace];
+      return [EthScopes.Mainnet];
```
danroc, line 51: Since this migration will be applied only to the accounts created by the example Snap, maybe we should migrate them to testnet instead.

packages/keyring-api/src/btc/types.test.ts
```diff
@@ -37,5 +53,25 @@ describe('types', () => {
         `${errorPrefix}: No separator character for ${address}`,
       );
     });
+
+    it('throws an error if there is multiple scopes', () => {
```
danroc, line 57:
```suggestion
    it('throws an error if there are multiple scopes', () => {
```

danroc (approved):

PR #161, opened 2025-01-27
refactor!: remove `HdKeyring` as default export

mikesposito:
After `@metamask/eth-hd-keyring` is migrated to Typescript, this PR edits its main file name and removes HdKeyeing as default export.

The number of line changes in the GitHub diff is misleading; the PR mostly renames `index.ts` into `hd-keyring.ts` and adds a new `index.ts` file which exports the `hd-keyring.ts` file contents

```
### Changed

- **BREAKING:** The `HdKeyring` is not exported as default anymore ([#161](https://github.com/MetaMask/accounts/pull/161))
```


packages/keyring-eth-hd/package.json
```diff
@@ -57,11 +57,11 @@
     "@lavamoat/preinstall-always-fail": "^2.1.0",
     "@metamask/auto-changelog": "^3.4.4",
     "@metamask/bip39": "^4.0.0",
-    "@metamask/eth-hd-keyring": "4.0.1",
     "@ts-bridge/cli": "^0.6.1",
     "@types/jest": "^29.5.12",
     "deepmerge": "^4.2.2",
-    "jest": "^29.5.0"
+    "jest": "^29.5.0",
+    "old-hd-keyring": "npm:@metamask/eth-hd-keyring@^4.0.1"
```
mikesposito, line 64: This package here was overriding the current version of `@metamask/eth-hd-keyring` confusing TS. With this line we can import the package with a different name

packages/keyring-eth-hd/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
```
mikesposito, line 10:
```suggestion
### Changed

```

packages/keyring-eth-hd/src/hd-keyring.ts
```diff
@@ -0,0 +1,590 @@
...
+   * @param withAccount - The address of the account.
+   * @param opts - The options for selecting the account.
+   * @returns The public key of the account.
+   */
+  async getEncryptionPublicKey(
+    withAccount: string,
+    opts: HDKeyringAccountSelectionOptions = {},
+  ): Promise<string> {
+    const privKey = this.#getPrivateKeyFor(withAccount, opts);
+    const publicKey = getEncryptionPublicKey(bytesToHex(privKey));
```
mikesposito, line 402: In the original implementation `privKey` was passed to `getEncryptionPublicKey` directly:
```typescript
const publicKey = getEncryptionPublicKey(privKey);
```
Though, the addition of types evidenced that `getEncryptionPublicKey` only accept hex strings, hence the conversion with `bytesToHex`

packages/keyring-eth-hd/src/hd-keyring.test.ts
mikesposito, line 1: There should be no functional change in these test cases, besides making TS happy

packages/keyring-eth-hd/package.json
mikesposito: The version `4.0.1` was pinned, but `yarn lint` was failing and I guess that `^4.0.1` serves the same purpose

packages/keyring-eth-hd/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from './hd-keyring';
```
cryptodev-2s, line 1: Making this change now makes it difficult to review unless we open it locally and compare. Could you separate this into two steps? First, focus only on the migration, and move the other changes to a second step. This way, we can concentrate solely on the types in this case.

mikesposito: I opened this other PR where there is only the migration (without the addition of `index.ts` and the rename to `hd-keyring.ts`): https://github.com/MetaMask/accounts/pull/166 
Unfortunately, as you can see, the file content still differs too much for Github to show it as a rename and edit. So I believe it doesn't bring many advantages to the review process

However, it can also be seen here in this PR at the specific commit where I migrated to TS: https://github.com/MetaMask/accounts/pull/161/commits/ee40095b27647f7b463334eb6d90a3d8614cf484

Though, let me know if you have further suggestions to show a better diff

mikesposito: Actually, there is[ this commit](https://github.com/MetaMask/accounts/pull/166/commits/bc89c460ecefb1334378cfc77ce87b8af2e0853a#diff-79137d999831949b43b4c2265b80e69a0f45626d482d2a0ab261241cabb44e68) on the new PR where the changes can be seen much more clearly. I will rebase this PR on top of the other one. Thanks for the suggestion

cryptodev-2s: can we use named exports instead of barrel export here as per our recent guidelines https://github.com/MetaMask/core/blob/main/docs/package-guidelines.md#guidelines-for-packages

mikesposito: Good idea! done in ae4fa8cd481103c1ae8c2c28f22ee11d1fd66686

gantunesr (approved):

PR #166, opened 2025-01-28
Migrate `HdKeyring` to typescript

mikesposito:
This is an alternative PR to #161 to migrate the keyring to typescript without renaming and moving the files according to the convention


packages/keyring-eth-hd/src/index.ts
```diff
@@ -0,0 +1,590 @@
...
+   * @param withAccount - The address of the account.
+   * @param opts - The options for selecting the account.
+   * @returns The public key of the account.
+   */
+  async getEncryptionPublicKey(
+    withAccount: string,
+    opts: HDKeyringAccountSelectionOptions = {},
+  ): Promise<string> {
+    const privKey = this.#getPrivateKeyFor(withAccount, opts);
+    const publicKey = getEncryptionPublicKey(bytesToHex(privKey));
```
mikesposito, line 402: In the original implementation `privKey` was passed to `getEncryptionPublicKey` directly:
```typescript
const publicKey = getEncryptionPublicKey(privKey);
```
Though, the addition of types evidenced that `getEncryptionPublicKey` only accept hex strings, hence the conversion with `bytesToHex`

packages/keyring-eth-hd/src/index.ts
```diff
@@ -0,0 +1,590 @@
...
+   * Get the wallet for the specified account and app origin.
+   *
+   * @param account - The address of the account.
+   * @param opts - The options for selecting the account.
+   * @returns A key pair representing the wallet.
+   */
+  #getWalletForAccount(
+    accounts: string,
+    opts: { withAppKeyOrigin: string },
+  ): { privateKey: Buffer; publicKey: Buffer };
```
mikesposito, line 511: This internal method is overloaded because the objects returned can vary significantly based on whether an app key origin is specified (i.e. a Buffer keypair is returned instead of an entire HDKey instance).

The actual signature of the implementation is left unchanged.

packages/keyring-eth-hd/package.json
```diff
@@ -57,11 +57,11 @@
     "@lavamoat/preinstall-always-fail": "^2.1.0",
     "@metamask/auto-changelog": "^3.4.4",
     "@metamask/bip39": "^4.0.0",
-    "@metamask/eth-hd-keyring": "4.0.1",
     "@ts-bridge/cli": "^0.6.1",
     "@types/jest": "^29.5.12",
     "deepmerge": "^4.2.2",
-    "jest": "^29.5.0"
+    "jest": "^29.5.0",
+    "old-hd-keyring": "npm:@metamask/eth-hd-keyring@^4.0.1"
```
mikesposito, line 64: This package used for testing backward compatibility was overwriting the current version of `@metamask/eth-hd-keyring`, causing TS to apply the same types of the new package. With this line we can import the package with a different name.

The version `4.0.1` was pinned, but `yarn lint` was failing and I guess that `^4.0.1` serves the same purpose

packages/keyring-eth-hd/test/index.test.ts
mikesposito, line 1: There should be no functional change in these test cases, besides the ones to make TS happy

packages/keyring-eth-hd/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **BREAKING**: This package now has no default export ([#161](https://github.com/MetaMask/accounts/pull/161))
```
mikesposito, line 12:
```suggestion
```

mikesposito:
@metamaskbot publish-preview

packages/keyring-eth-hd/src/index.ts
```diff
@@ -123,187 +168,423 @@ class HdKeyring {
...
     assertIsHexString(data);
     const message = remove0x(data);
-    const privKey = this._getPrivateKeyFor(address, opts);
-    const msgSig = ecsign(Buffer.from(message, 'hex'), privKey);
-    const rawMsgSig = concatSig(msgSig.v, msgSig.r, msgSig.s);
+    const privKey = this.#getPrivateKeyFor(address, opts);
+    const msgSig = ecsign(Buffer.from(message, 'hex'), Buffer.from(privKey));
+    const rawMsgSig = concatSig(
+      // WARN: verify this cast to Buffer
+      msgSig.v as unknown as Buffer,
```
mikesposito, line 292: It looks like `@metamask/eth-sig-util` only accepts a Buffer here, while `msgSig.v` seems to be a `bigint`

packages/keyring-eth-hd/test/index.test.ts
```diff
@@ -154,7 +159,7 @@ describe('hd-keyring', () => {
         numberOfAccounts: 2,
       });
 
-      const accounts = await keyring.getAccounts();
+      const accounts = keyring.getAccounts();
```
mikesposito, line 162: `getAccounts()` was awaited everywhere in the tests, but the function implementation was not marked as `async`

mikesposito: we can change the tests here, and then change them back along with fixing the implementation when `HDKeyring` will implement the `Keyring` type

packages/keyring-eth-hd/src/index.ts
mcmire: Hmm, does that cause an issue? A bigint is not a Buffer.

packages/keyring-eth-hd/src/index.ts
```diff
@@ -0,0 +1,593 @@
...
+  }
+
+  /**
+   * Get the addresses of all accounts in the keyring.
+   *
+   * @returns The addresses of all accounts in the keyring.
+   */
+  getAccounts(): string[] {
+    return this.#wallets.map((wallet) => {
+      assert(wallet.publicKey, 'Expected public key to be set');
```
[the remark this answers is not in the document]
mikesposito: Hmm, this `assert` comes from `@metamask/utils` instead of the built-in Node.js' `assert`. Looking at its implementation it seems to be compatible with a browser environment as well: https://github.com/MetaMask/utils/blob/90837bad1d85277aa5b8f16fe40beb81e0a1e09d/src/assert.ts#L81

Having this in mind, do you still think we should change these to use an explicit _if undefined_ ?

packages/keyring-eth-hd/src/index.ts
```diff
@@ -0,0 +1,593 @@
...
+    } = { version: SignTypedDataVersion.V1 },
+  ): Promise<string> {
+    // Treat invalid versions as "V1"
+    const version = Object.keys(SignTypedDataVersion).includes(opts.version)
+      ? opts.version
+      : SignTypedDataVersion.V1;
+
+    const privateKey = this.#getPrivateKeyFor(withAccount, opts);
+    return signTypedData({
+      privateKey: Buffer.from(privateKey),
```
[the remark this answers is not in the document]
mikesposito: Since `#getPrivateKeyFor` returns a Uint8Array or a Buffer depending on whether we are passing `{ withAppKeyOrigin: string }` as `opts`, `signTypedData` could receive something that is not a Buffer.

using `Buffer.from(privateKey)` should convert the Uint8Array to a Buffer or just return another Buffer with the same content as the original `privateKey`

packages/keyring-eth-hd/src/index.ts
```diff
@@ -0,0 +1,593 @@
...
+   * Convert a mnemonic to a Uint8Array mnemonic.
+   *
+   * @param mnemonic - The mnemonic seed phrase.
+   * @returns The Uint8Array mnemonic.
+   */
+  #mnemonicToUint8Array(
+    mnemonic: Buffer | JsCastedBuffer | string | Uint8Array | number[],
+  ): Uint8Array {
+    let mnemonicData: unknown = mnemonic;
+    // when encrypted/decrypted, buffers get cast into js object with a property type set to buffer
```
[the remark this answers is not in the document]
mikesposito: The comment is left unchanged from the original repo, though I think that it is referring to the output of `buffer.toJSON()` which returns an object with this shape:
```
{
  type: 'Buffer',
  data: [
     71, 101, 101, 107,
    115, 102, 111, 114,
     71, 101, 101, 107,
    115
  ]
}
```

Perhaps we can change it to be something like this:
```suggestion
    // When using Buffer.toJSON(), the Buffer is serialized into an object
    // with the structure `{ type: 'Buffer', data: [...] }`
```

mikesposito:
The following extension PR is using a package preview from this PR, which appears to be working fine: https://github.com/MetaMask/metamask-extension/pull/29961

packages/keyring-eth-hd/src/index.ts
```diff
@@ -0,0 +1,593 @@
...
+    this.root = this.hdWallet.derive(this.hdPath);
+  }
+
+  /**
+   * Get the address of the account from the public key.
+   *
+   * @param publicKey - The public key of the account.
+   * @returns The address of the account.
+   */
+  #addressfromPublicKey(publicKey: Uint8Array): string {
```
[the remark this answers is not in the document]
mikesposito: I changed all types related to addresses from `string` to `Hex`. I guess you were referring to `add0x` which is now used to normalize addresses when needed (added `HdKeyring.#normalizeAddress`)

packages/keyring-eth-hd/src/index.ts
mcmire: Ah, strange, okay.

packages/keyring-eth-hd/src/index.ts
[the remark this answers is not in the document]
mcmire: Ah, Sorry I didn't realize where this came from. No, this is okay.

packages/keyring-eth-hd/src/index.ts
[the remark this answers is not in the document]
mcmire: Oh yeah! I like that.

mcmire (dismissed):
Looks good to me! Nice work :)

mcmire (approved):
LGTM!

yarn.lock
```diff
@@ -9166,6 +9153,19 @@ __metadata:
   languageName: node
   linkType: hard
 
+"old-hd-keyring@npm:@metamask/eth-hd-keyring@^4.0.1":
+  version: 4.0.2
+  resolution: "@metamask/eth-hd-keyring@npm:4.0.2"
```
ccharly, line 9158: We are not longer testing the `4.0.1`, probably because of the `^` now.

What was the `yarn lint` issue you got? Was it caused by syncpack?

gantunesr (approved):

yarn.lock
ccharly: As discussed internally, that should be ok to use the `4.0.2` for our test. 👍

PR #175, opened 2025-01-30
feat: add dedicated OneKey keyring class

Akaryatrh:
This PR adds support for a dedicated OneKey keyring (until now it was sharing the same keyring instance than Trezor) so it's considered as a standalone device and could get its own tag inside account list.

There are two others PRs:

- metamask-extension: https://github.com/MetaMask/metamask-extension/pull/29999
- core (accounts and keyring controllers): https://github.com/MetaMask/core/pull/5216

Fixes: https://github.com/MetaMask/accounts-planning/issues/793

### DIFF
--- a/packages/keyring-eth-trezor/src/onekey-keyring.test.ts
+++ b/packages/keyring-eth-trezor/src/onekey-keyring.test.ts
@@ -1,0 +1,42 @@
+import HDKey from 'hdkey';
+import * as sinon from 'sinon';
+
+import { OneKeyKeyring } from './onekey-keyring';
+import { TrezorBridge } from './trezor-bridge';
+import { TrezorKeyring } from './trezor-keyring';
+
+const fakeXPubKey =
+  'xpub6FnCn6nSzZAw5Tw7cgR9bi15UV96gLZhjDstkXXxvCLsUXBGXPdSnLFbdpq8p9HmGsApME5hQTZ3emM2rnY5agb9rXpVGyy3bdW6EEgAtqt';
+const fakeHdKey = HDKey.fromExtendedKey(fakeXPubKey);
+
+describe('TrezorKeyring', function () {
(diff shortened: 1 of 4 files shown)


packages/keyring-eth-trezor/src/onekey-keyring.ts
```diff
@@ -0,0 +1,9 @@
+import { TrezorKeyring } from './trezor-keyring';
+
+const oneKeyKeyringType = 'OneKey Hardware';
+
+export class OneKeyKeyring extends TrezorKeyring {
```
ccharly, line 5: This should also be exported in `src/index.ts` maybe?

Akaryatrh: Woops, indeed! Fixed 🙂

packages/keyring-eth-trezor/src/onekey-keyring.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import { OneKeyKeyring } from './onekey-keyring';
+import { TrezorBridge } from './trezor-bridge';
+import { TrezorKeyring } from './trezor-keyring';
+
+const fakeXPubKey =
+  'xpub6FnCn6nSzZAw5Tw7cgR9bi15UV96gLZhjDstkXXxvCLsUXBGXPdSnLFbdpq8p9HmGsApME5hQTZ3emM2rnY5agb9rXpVGyy3bdW6EEgAtqt';
+const fakeHdKey = HDKey.fromExtendedKey(fakeXPubKey);
+
+describe('TrezorKeyring', function () {
```
ccharly, line 12:
```suggestion
describe('OneKeyKeyring', function () {
```

ccharly (approved):

PR #187, opened 2025-02-03
fix(29917): Trezor accounts not able to add multiple HD path accounts into the account list

Akaryatrh:
Currently it's not possible to add accounts from different HDPaths with Trezor related devices. For some reason, on a path change we do the exact same operations than when we "forget" a device. This is not something that we do for ledger devices so it appears useless (and then creates bugs).

Fixes https://github.com/MetaMask/metamask-extension/issues/29917

### DIFF
--- a/packages/keyring-eth-trezor/src/trezor-keyring.test.ts
+++ b/packages/keyring-eth-trezor/src/trezor-keyring.test.ts
@@ -133,6 +133,15 @@ describe('TrezorKeyring', function () {
     });
   });
 
+  describe('getModel', function () {
+    it('gets bridge model', async function () {
+      keyring.bridge.model = 'foo';
+      const model = keyring.getModel();
+
+      expect(model).toBe('foo');
+    });
+  });
+
   describe('init', function () {
     it('initialises the bridge', async function () {
       const initStub = sinon.stub().resolves();
--- a/packages/keyring-eth-trezor/src/trezor-keyring.test.ts
+++ b/packages/keyring-eth-trezor/src/trezor-keyring.test.ts
@@ -729,10 +729,9 @@ describe('TrezorKeyring', function () {
       expect(keyring.paths).toStrictEqual(mockPaths);
     });
 
-    it('should update the hdPath and reset account and page properties if passed a new hdPath', async function () {
+    it('should update the hdPath if passed a new hdPath', async function () {
       const SLIP0044TestnetPath = `m/44'/1'/0'/0`;
       keyring.setHdPath(SLIP0044TestnetPath);
-
       expect(keyring.hdPath).toBe(SLIP0044TestnetPath);
       expect(keyring.hdk.publicKey).toBeNull();
     });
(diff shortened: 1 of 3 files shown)


Akaryatrh:
@metamaskbot publish-preview

packages/keyring-eth-trezor/src/trezor-keyring.ts
```diff
@@ -539,11 +539,6 @@ export class TrezorKeyring extends EventEmitter {
     // Reset HDKey if the path changes
     if (this.hdPath !== hdPath) {
       this.hdk = new HDKey();
```
mikesposito, line 541: Curious: why the `HDKey` instance needs to be reset?

mikesposito: Oh, maybe because `this.hdk` is initialized with the public key related to the selected path

packages/keyring-eth-trezor/src/trezor-keyring.ts
```diff
@@ -539,11 +539,6 @@ export class TrezorKeyring extends EventEmitter {
...
-      this.accounts = [];
-      this.page = 0;
-      this.perPage = 5;
-      this.unlockedAccount = 0;
-      this.paths = {};
```
mikesposito, line 546: Will the keyring still be able to sign transactions with the accounts previously derived with a different path? Since `this.hdk` is specific to a path, after changing path we won't have the original public key used to derive the first accounts

packages/keyring-eth-trezor/src/trezor-keyring.test.ts
```diff
@@ -133,6 +133,15 @@ describe('TrezorKeyring', function () {
...
 
+  describe('getModel', function () {
+    it('gets bridge model', async function () {
+      keyring.bridge.model = 'foo';
+      const model = keyring.getModel();
+
+      expect(model).toBe('foo');
+    });
+  });
+
```
mikesposito, line 144: is this related to this PR?

Akaryatrh: Yes and no, I had to increase the missing coverage percentage (that went down after modifying `trezor-keyring.test.ts` file)

packages/keyring-eth-trezor/src/trezor-keyring.ts
Akaryatrh: Indeed, previously derived accounts are not able to sign or send tx. I guess i need to check what we're doing with Ledger keyring as it's working there.

mikesposito: I think it's warranted to add a test case for this scenario as well

mikesposito: the LedgerKeyring stores a map of [`AccountDetails`](https://github.com/MetaMask/accounts/blob/dd7b8df799787b0cd178c971e7fbc75660792c27/packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts#L45) for each derived address, so even if the path is switched it can be retrieved again when signing: https://github.com/MetaMask/accounts/blob/dd7b8df799787b0cd178c971e7fbc75660792c27/packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts#L470

PR #189, opened 2025-02-03
feat: add Solana methods from wallet-standard

aganglada:
This change adds support for Solana wallet standard methods and removes the previous ones


danroc:
@aganglada, is the `solana:` prefix present in the requests sent by dapps? If not, the RPC Router won't be able to send them to the right Snap.

PR #191, opened 2025-02-05
feat: document supported Solana methods

aganglada:
Adds Solana methods definition


packages/keyring-api/docs/sol-methods.md
```diff
@@ -0,0 +1,481 @@
...
+## [solana:signAndSendTransaction](https://github.com/anza-xyz/wallet-standard/blob/master/packages/core/features/src/signAndSendTransaction.ts#L10)
+
+Signs and sends a transaction to the Solana blockchain.
+
+### [Parameters](https://github.com/anza-xyz/wallet-standard/blob/master/packages/core/features/src/signAndSendTransaction.ts#L42)
+
+- **Transaction intent (required)**
+  - Type: `object`
+  - Properties:
+    - `account` (required)
```
aganglada, line 15: do we need all the properties inside the account?

packages/keyring-api/docs/sol-methods.md
```diff
@@ -0,0 +1,481 @@
...
+
+```json
+[
+  {
+    "signature": "1234567890"
+  }
+]
+```
+
+## [solana:signIn](https://github.com/anza-xyz/wallet-standard/blob/master/packages/core/features/src/signIn.ts#L4)
```
aganglada, line 389: Do we need to implement this method?

packages/keyring-api/docs/sol-methods.md
```diff
@@ -0,0 +1,481 @@
...
+      - Description: The transaction signature.
+      - Type: `string`
+
+### Examples
+
+**Request:**
+
+```json
+{
+  "method": "solana:signAndSendTransaction",
```
vandan, line 74: While it does precisely match wallet-standard, the "solana:" prefix seems unnecessary given that the method will only be available through a solana-specific scope in the Multichain API anyways. The namespacing of method names using prefixes is more an artifact of the way wallet-standard attempts to solve multichain calls. The multichain API doesn't require such namespacing in method names.

packages/keyring-api/docs/sol-methods.md
danroc: IMO the address (or the public key) should be enough

packages/keyring-api/docs/sol-methods.md
```diff
@@ -0,0 +1,481 @@
...
+          - Type: `string`
+        - `icon` (optional)
+          - Description: The icon of the account.
+          - Type: `string`
+    - `transaction` (required)
+      - Description: The base64 encoded transaction to sign and send.
+      - Type: `string`
+    - `chain` (required)
+      - Description: The chain ID of the transaction.
+    - [Type](https://github.com/wallet-standard/wallet-standard/blob/c6fa5fd7d58e9ea1e6762127d16585fbb56ff88a/packages/core/base/src/identifier.ts#L8): `${namespace}:${reference}`
```
danroc, line 50:
```suggestion
      - [Type](https://github.com/wallet-standard/wallet-standard/blob/c6fa5fd7d58e9ea1e6762127d16585fbb56ff88a/packages/core/base/src/identifier.ts#L8): `${namespace}:${reference}`
```

packages/keyring-api/docs/sol-methods.md
aganglada: @vandan what's your take on this?

packages/keyring-api/docs/sol-methods.md
jiexi: Probably should for better perceived solona support at launch. Doesn't seem like there are fall back account features for this method either

https://github.com/anza-xyz/wallet-standard/blob/6cfb9a0be49351dba765131e58d57536c31ae251/packages/wallet-adapter/base/src/adapter.ts#L459-L479

packages/keyring-api/docs/sol-methods.md
jiexi: Agreed with Vandan. If the methods formatted this way at this level of code will result in also having to also reference them with the same solona prefix at the API level, that would extremely unideal. If there is some translation/transformation happening in between these two points, then I'm fine with any internal representation that works better for you guys. 

ideally the wallet_invoke method call for solona does NOT look like this
```
{
    type: 'caip-x',
    data: {
        "jsonrpc": "2.0",
        method: 'wallet_invokeMethod',
        params: {
            scope: 'solona:mainnet',
            request: {
                "method": "solona:signAndSendTransaction",
                "params": {...},
            }
        }
    }
}

```

packages/keyring-api/docs/sol-methods.md
jiexi: I'm not entirely sure, but the wallet-standard wallet adapter seems to want these fields. I don't know what the plan is with how we're going to implement wallet-standard though and if these are necessary for our implementation. If this is the method signature being exposed on the dapp facing API, it does indeed seem odd to have many of these other fields in this `account` object.

packages/keyring-api/docs/sol-methods.md
```diff
@@ -0,0 +1,481 @@
+# SOL Methods
+
+Here we document the Solana methods that an account Snap may implement to
+support requests originated from dapps.
+
+## [solana:signAndSendTransaction](https://github.com/anza-xyz/wallet-standard/blob/master/packages/core/features/src/signAndSendTransaction.ts#L10)
+
+Signs and sends a transaction to the Solana blockchain.
+
+### [Parameters](https://github.com/anza-xyz/wallet-standard/blob/master/packages/core/features/src/signAndSendTransaction.ts#L42)
```
aganglada, line 10: Based on the wallet standard, how would we know the origin of the request?

packages/keyring-api/docs/sol-methods.md
EdouardBougon: I agree. We need this information on the wallet-standard implementation side to add this kind of [checks](https://github.com/onemoongate/moongate-adapter/blob/fe611be09b3d8fc18a73575084e7e8f7deb5d03d/src/wallet.ts#L342).

But I don't think it's relevant on snap side.

packages/keyring-api/docs/sol-methods.md
```diff
@@ -0,0 +1,481 @@
...
+            - Type: `string`
+            - [Type](https://github.com/wallet-standard/wallet-standard/blob/c6fa5fd7d58e9ea1e6762127d16585fbb56ff88a/packages/core/base/src/identifier.ts#L8): `${namespace}:${reference}`
+            - Example: `solana:signAndSendTransaction`
+        - `label` (optional)
+          - Description: The label of the account.
+          - Type: `string`
+        - `icon` (optional)
+          - Description: The icon of the account.
+          - Type: `string`
+    - `transaction` (required)
```
EdouardBougon, line 45: Each of the following method could accept an array of input:
- `signTransaction` (See this [line](https://github.com/anza-xyz/wallet-standard/blob/6cfb9a0be49351dba765131e58d57536c31ae251/packages/core/features/src/signTransaction.ts#L35))
- `signAndSendTransaction` (See this [line](https://github.com/anza-xyz/wallet-standard/blob/6cfb9a0be49351dba765131e58d57536c31ae251/packages/core/features/src/signAndSendTransaction.ts#L38))
- `signMessage` (See this [line](https://github.com/anza-xyz/wallet-standard/blob/6cfb9a0be49351dba765131e58d57536c31ae251/packages/core/features/src/signMessage.ts#L23))

The difference between `signAndSendTransaction` with an array on input and `signAndSendAllTransactions` is [this feature](https://github.com/anza-xyz/wallet-standard/blob/6cfb9a0be49351dba765131e58d57536c31ae251/packages/core/features/src/signAndSendAllTransactions.ts#L48).

Even though Phantom and the Solflare wallet support this feature, the snap doesn’t need to in the first version. In [Moongate’s implementation](https://github.com/onemoongate/moongate-adapter/blob/fe611be09b3d8fc18a73575084e7e8f7deb5d03d/src/wallet.ts#L380), the wallet iterates over the input array for the `signMessage`.

packages/keyring-api/docs/sol-methods.md
jiexi: I was thinking this would be available in the request context, much in the same way it is EVM

packages/keyring-api/docs/sol-methods.md
vandan: Is it really a separate method for the Snap though?
In EVM-land Sign-in with Ethereum is just a different way of calling `personal_sign` with a payload that get's slightly different handling in our signature/confirmation UI. I'm wondering if it's similar for Solana where SignIn and regular Signatures can be supported through one `signMessage` method.

packages/keyring-api/docs/sol-methods.md
aganglada: Removing everything except the address field

packages/keyring-api/docs/sol-methods.md
aganglada: @EdouardBougon do you know if this is commonly used by Dapps?
We don't have a UI to support this type of request at the moment.

EdouardBougon: I don’t have an example of a Dapp using it, but I created a demo to simulate it with the three main Solana wallets.
For example, on Phantom, the UI is the same.

https://docs.google.com/presentation/d/1-EZesZXLd9f2T2And5u8KMJ6UvqO3mIGl5AzmzkxNbA/edit?usp=sharing

For V1, we can handle this on the Wallet Standard side and iterate on each message to sign them one by one on MetaMask.

packages/keyring-api/docs/sol-methods.md
aganglada: Added

packages/keyring-api/docs/sol-methods.md
```diff
@@ -0,0 +1,401 @@
...
+## [signAndSendTransaction](https://github.com/anza-xyz/wallet-standard/blob/master/packages/core/features/src/signAndSendTransaction.ts#L10)
+
+Signs and sends a transaction to the Solana blockchain.
+
+### [Parameters](https://github.com/anza-xyz/wallet-standard/blob/master/packages/core/features/src/signAndSendTransaction.ts#L42)
+
+- **Transaction intent (required)**
+  - Type: `object`
+  - Properties:
+    - `origin` (required)
```
aganglada, line 15: Team, if the origin if MetaMask, should we send `metamask.io`? or should we send a `metamask` string

danroc: The MetaMask origin should be `metamask`, it's the origin we use elsewhere.

But it shouldn't be part of the signing request itself because it needs to be set by the **client** in a chain-agnostic way. I think it should be added to the [`KeyringRequest`](https://github.com/MetaMask/accounts/blob/main/packages/keyring-api/src/api/request.ts) type.

aganglada: so next to the nested `request`?

aganglada: Removed, then to be added on the `KeyringRequest`

jiexi (dismissed):

packages/keyring-api/docs/sol-methods.md
```diff
@@ -0,0 +1,381 @@
...
+
+```json
+{
+  "signature": "1234567890",
+  "signedMessage": "1234567890",
+  "signatureType": "ed25519"
+}
+```
+
+## [signAndSendAllTransactions](https://github.com/anza-xyz/wallet-standard/blob/master/packages/core/features/src/signAndSendAllTransactions.ts#L8)
```
EdouardBougon, line 205: I got confirmation that Phantom does not handle this feature.
They handle the signing of a transaction array.

But not the signing and sending of a transaction array. In this case, they iterate over the array and call `signAndSendTransaction`.

So, in the end, we don’t need this feature.

packages/keyring-api/docs/sol-methods.md
EdouardBougon: I got confirmation that Phantom iterates on each message and calls `signMessage`, and it does the same for `signAndSendTransaction`.

Only `signTransaction` can manage an array of transactions on Phantom’s side. But as mentioned, we can iterate on each transaction on our side for the first version.

danroc (approved):

Issue #143, opened 2025-01-15
`LedgerKeyring` class from `@metamask/eth-ledger-bridge-keyring` should implement the `Keyring` type


mikesposito:
All keyring classes should extend from [this base common type](https://github.com/MetaMask/utils/blob/main/src/keyring.ts).

PR #194, opened 2025-02-05
refactor!: `LedgerKeyring` implements `Keyring` type

mikesposito:
The `LedgerKeyring` is being overhauled to implement the `Keyring` type from `@metamask/keyring-utils`. Most of the changes are due to the addresses type that is now `Hex` 

### Changed

- **BREAKING:** `LedgerKeyring` now implements the `Keyring` type ([#194](https://github.com/MetaMask/accounts/pull/194))
  - The class does not extend `EventEmitter` anymore.
  - The `LedgerKeyring.accounts` class variable is now a `readonly Hex[]` array.
  - The `addAccounts` method signature has been changed
    - An `amount` number parameter is now required to specify the number of accounts to add.
    - The method now returns a promise resolving to an array of `Hex` addresses.
  - The `unlock` method now returns `Promise<Hex>`
  - The `getAccounts` method now returns `Promise<Hex[]>`
  - The `deserialize` method now requires a `LedgerKeyringSerializedState` typed parameter
  - The `signTransaction` method now accepts an `Hex` typed value as the `address` parameter
  - The `signMessage` method now accepts an `Hex` typed value as the `withAccount` parameter
  - The `signPersonalMessage` method now accepts an `Hex` typed value as the `withAccount` parameter
  - The `signTypedData` method now accepts an `Hex` typed value as the `withAccount` parameter
  - The `unlockAccountByAddress` method now accepts an `Hex` typed value as the `address` parameter

### Removed

- **BREAKING:** The `exportAccount` method has been removed

---
* Fixes: https://github.com/MetaMask/accounts/issues/143


packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -211,9 +209,11 @@ export class LedgerKeyring extends EventEmitter {
     this.hdPath = hdPath;
   }
 
-  async unlock(hdPath?: string, updateHdk = true): Promise<string> {
+  async unlock(hdPath?: string, updateHdk = true): Promise<Hex> {
     if (this.isUnlocked() && !hdPath) {
-      return 'already unlocked';
+      return this.#getChecksumHexAddress(
+        ethUtil.publicToAddress(this.hdk.publicKey, true).toString('hex'),
+      );
```
mikesposito, line 216: I'm not sure how to handle this case here. The original code returned `already unlocked` if:
- there is already a `this.hdk` defined
- `hdPath` is not passed to the function (i.e. the consumer calls `keyring.unlock()`

However, since we want this function to return a value of type `Hex` we must return a valid address. Based on the lines below the actual `publicKey` returned from this function comes from the device (`payload.publicKey`) and it's used to construct an HDKey to be stored in `this.hdk`. So, the new code assumes that we want to return the address derived from the public key stored in `this.hdk.publicKey`

mikesposito:
Marking the PR as draft again because a new `@metamask/utils` major version is being prepared for release, which will ship breaking changes to the `Keyring` type

packages/keyring-eth-ledger-bridge/CHANGELOG.md
```diff
@@ -7,6 +7,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+    - An `amount` number parameter is now required to specify the number of accounts to add.
+    - The method now returns a promise resolving to an array of `Hex` addresses.
+  - The `unlock` method now returns `Promise<Hex>`
+  - The `getAccounts` method now returns `Promise<Hex[]>`
+  - The `deserialize` method now requires a `LedgerKeyringSerializedState` typed parameter
+  - The `signTransaction` method now accepts an `Hex` typed value as the `address` parameter
+  - The `signMessage` method now accepts an `Hex` typed value as the `withAccount` parameter
+  - The `signPersonalMessage` method now accepts an `Hex` typed value as the `withAccount` parameter
+  - The `signTypedData` method now accepts an `Hex` typed value as the `withAccount` parameter
+  - The `unlockAccountByAddress` method now accepts an `Hex` typed value as the `address` parameter
```
ccharly, line 25:
```suggestion
  - The `unlock` method now returns `Promise<Hex>`.
  - The `getAccounts` method now returns `Promise<Hex[]>`.
  - The `deserialize` method now requires a `LedgerKeyringSerializedState` typed parameter.
  - The `signTransaction` method now accepts an `Hex` typed value as the `address` parameter.
  - The `signMessage` method now accepts an `Hex` typed value as the `withAccount` parameter.
  - The `signPersonalMessage` method now accepts an `Hex` typed value as the `withAccount` parameter.
  - The `signTypedData` method now accepts an `Hex` typed value as the `withAccount` parameter.
  - The `unlockAccountByAddress` method now accepts an `Hex` typed value as the `address` parameter.
```

packages/keyring-eth-ledger-bridge/CHANGELOG.md
```diff
@@ -7,6 +7,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Removed
+
+- **BREAKING:** The `exportAccount` method has been removed
```
ccharly, line 29:
```suggestion
- **BREAKING:** The `exportAccount` method has been removed ([#194](https://github.com/MetaMask/accounts/pull/194))
```

packages/keyring-eth-ledger-bridge/src/ledger-keyring.test.ts
```diff
@@ -95,17 +96,17 @@ describe('LedgerKeyring', function () {
    */
   async function basicSetupToUnlockOneAccount(accountIndex = 0): Promise<void> {
     keyring.setAccountToUnlock(accountIndex);
-    await keyring.addAccounts();
+    await keyring.addAccounts(1);
     jest
       .spyOn(keyring, 'unlock')
-      .mockResolvedValue(fakeAccounts[accountIndex] as string);
+      .mockResolvedValue(fakeAccounts[accountIndex] as Hex);
```
ccharly, line 102: Very nit: I wonder if we could make `fakeAccounts` a `Hex[]` instead? That would avoid type-casting when ever we use them in the test (we were already type-casting, so I'm ok keeping this as-is too).

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -211,9 +210,11 @@ export class LedgerKeyring extends EventEmitter {
     this.hdPath = hdPath;
   }
 
-  async unlock(hdPath?: string, updateHdk = true): Promise<string> {
+  async unlock(hdPath?: string, updateHdk = true): Promise<Hex> {
     if (this.isUnlocked() && !hdPath) {
-      return 'already unlocked';
+      return this.#getChecksumHexAddress(
+        ethUtil.publicToAddress(this.hdk.publicKey, true).toString('hex'),
```
ccharly, line 216: Nit: What about using the same pattern for `ethUtil` than the Trezor's PR, where we import `publicToAddress`, etc.. ?

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
ccharly: Your new logic seems fine (and way better) to me. IDK if we were using the `already unlocked` value somewhere else, but that logic was kinda "fragile" IMO.

ccharly: I would just add a comment explaining this maybe? Like we re-use the already derived public key in that case (not sure how to phrase that though..)

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -459,7 +460,7 @@ export class LedgerKeyring extends EventEmitter {
     return signature;
   }
 
-  async unlockAccountByAddress(address: string): Promise<string | undefined> {
+  async unlockAccountByAddress(address: Hex): Promise<string | undefined> {
     const checksummedAddress = ethUtil.toChecksumAddress(address);
```
ccharly, line 464: Maybe we could use `getChecksumAddress` instead? (similarly to the Trezor's PR)

mikesposito: Good idea. Done in [6b04afc](https://github.com/MetaMask/accounts/pull/194/commits/6b04afc00591c9ef35d2403940963c27d69845e2)

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
mikesposito: Done in [6b04afc](https://github.com/MetaMask/accounts/pull/194/commits/6b04afc00591c9ef35d2403940963c27d69845e2)

packages/keyring-eth-ledger-bridge/src/ledger-keyring.test.ts
mikesposito: If we use `fakeAccounts: Hex[]` instead of `fakeAccounts = [..] as const` (readonly constant) then we'll have to non-null assert (`!`) in all places where we access the array with an arbitrary index

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
mikesposito: Definitely! done in [aa850a3](https://github.com/MetaMask/accounts/pull/194/commits/aa850a3a7eb58790f6b0f7eb1f19328d6f1a7979)

packages/keyring-eth-ledger-bridge/src/ledger-keyring.test.ts
ccharly: I was not aware that `as const` would statically enforce the array static content (which makes total sense actually, thanks for sharing this 😄).

ccharly (approved):

ccharly (approved):

PR #207, opened 2025-02-13
feat(keyring-snap-bridge): make account creation async

ccharly:
We now resume the Snap execution before persisting the account into the Snap keyring state.

This allows the Snap to have an existing account (once approved or not by the user, just like before), but this delays the actual account creation on the MetaMask clients (more specifically, the `AccountsController` in this case), which allows other controllers to react to the `AccountsController:accountCreated` event and to directly fetch account informations (assets/balances) from the Snap.

In the following example, the `SomeOtherController` cannot use the account because the account wasn't persisted by the Snap yet:
```ts
// Snap keyring methods
snap.createAccount = () => {
  const account = ...;
  
  await emitSnapKeyringEvent(AccountCreated, { account }); // -> This will trigger keyring.handleSnapkeyringEvent
  
  // Now persists the account (since we know MetaMask has accepted it).
  state.accounts.push(account);
}

// Snap keyring (on the MetaMask client)
keyring.handleSnapKeyringEvent = (message) => {
  ...
  switch (message.method) {
    case 'AccountCreated': {
      // Show some dialogs for the user
      if (dialog.accepted) {
        // At this point, Snap execution is still pending...
        await keyring.saveState(); // This will trigger AccountsController:accountAdded (see addNewAccount method in this example) event 
      }
    }
  ...
})

AccountsController.addNewAccount = (account) => {
  this.messenger.publish(`AccountsController:accountAdded`, account); // This will trigger SomeOtherController.onAccountAdded handler
}

SomeOtherController.onAccountAdded = (account) => {
  // !
  // ! Here was the problem, since the account was still not persisted onto the Snap.
  // !
  await this.getSnapClient(account.metadata.snap.id).listAccountAssets();
}
```

### DIFF
--- a/packages/keyring-snap-bridge/src/SnapKeyring.ts
+++ b/packages/keyring-snap-bridge/src/SnapKeyring.ts
@@ -207,6 +207,11 @@ export class SnapKeyring extends EventEmitter {
       displayConfirmation,
     } = message.params;
 
+    // READ THIS CAREFULLY:
+    // ------------------------------------------------------------------------
+    // The account creation flow is now asynchronous. We expect the Snap to
+    // first create its account data and then fire the "AccountCreated" event.
+
     // Potentially migrate the account.
     const account = transformAccount(newAccountFromEvent);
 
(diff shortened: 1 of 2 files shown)


ccharly:
@metamaskbot publish-previews

ccharly:
@metamaskbot publish-previews

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -207,6 +207,11 @@ export class SnapKeyring extends EventEmitter {
       displayConfirmation,
     } = message.params;
 
+    // READ THIS CAREFULLY:
+    // ------------------------------------------------------------------------
+    // The account creation flow is now asynchronous. We expect the Snap to
+    // first create its account data and then fire the "AccountCreated" event.
```
danroc, line 213:
```suggestion
    // first create the account data and then fire the "AccountCreated" event.
```

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -224,18 +229,40 @@ export class SnapKeyring extends EventEmitter {
       throw new Error(`Account '${account.id}' already exists`);
     }
 
+    // Add the account on the keyring, but wait for the MetaMask client to first
+    // approve this first.
```
danroc, line 233:
```suggestion
    // Add the account to the keyring, but wait for the MetaMask client to
    // approve the account creation first.
```

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -224,18 +229,40 @@ export class SnapKeyring extends EventEmitter {
...
     await this.#callbacks.addAccount(
       address,
       snapId,
+      // This callback is passed to the MetaMask client, it will be called whenever
+      // the end user will accept or not the account creation.
       async (accepted: boolean) => {
         if (accepted) {
+          // We only consider the account to be created on the Snap keyring only if
```
danroc, line 241:
```suggestion
          // We consider the account to be created on the Snap keyring only if
```

danroc (approved):

PR #208, opened 2025-02-13
fix: export `keyring` from `keyring-utils`

mikesposito:
With https://github.com/MetaMask/accounts/pull/201 the `Keyring` type has been added to `@metamask/keyring-utils`, specifically under `src/keyring.ts`. Though this new file is currently not exported from `index.ts`. This PR fixes that


ccharly (approved):
🙈

PR #209, opened 2025-02-14
chore: upgrade ethereumjs packages

matthewwalsh0:
Upgrade all `@ethereumjs/*` packages:

- `@ethereumjs/util` - `^8.1.0` > `^9.1.0`
- `@ethereumjs/tx` - `^4.2.0` > `^5.4.0`
- `@ethereumjs/common` - `^3.2.0` > `^4.4.0`

Necessary to align with `@metamask/transaction-controller` which requires the latest versions to support EIP-7702 transactions.

Minor refactors to support breaking changes, primarily:

- `Buffer` usage replaced with `Uint8Array`.
- `bufferToHex` > `bytesToHex`
- `Transaction` Constructor > `LegacyTransaction.fromTxData`
- `TxData` > `TypedTxData`
- `getMessageToSign(false)` > `getMessageToSign`
- `getMessageToSign(true)` > `getHashedMessageToSign`

Notes

- Using `bytesToHex` from `@metamask/utils` rather than `@ethereumjs/utils` for additional control and performance.
- Intentionally doing the minimum refactors to private methods and existing logic hence some continued use of `Buffer`.


matthewwalsh0:
@metamaskbot publish-preview

matthewwalsh0:
@metamaskbot publish-preview

packages/keyring-eth-hd/src/hd-keyring.ts
```diff
@@ -253,7 +248,7 @@ export class HdKeyring {
       privateKey instanceof Uint8Array,
       'Expected private key to be of type Uint8Array',
     );
-    return bytesToHex(privateKey);
+    return remove0x(bytesToHex(privateKey));
```
danroc, line 251: Is this expected to be a breaking change?

packages/keyring-eth-hd/src/hd-keyring.ts
```diff
@@ -425,7 +419,7 @@ export class HdKeyring {
     opts: HDKeyringAccountSelectionOptions = {},
   ): Promise<string> {
     const privKey = this.#getPrivateKeyFor(withAccount, opts);
-    const publicKey = getEncryptionPublicKey(bytesToHex(privKey));
+    const publicKey = getEncryptionPublicKey(remove0x(bytesToHex(privKey)));
```
danroc, line 422: Same question here ^

packages/keyring-eth-hd/src/hd-keyring.ts
```diff
@@ -567,8 +561,8 @@ export class HdKeyring {
       assert(privateKey, 'Expected private key to be set');
       const appKeyOriginBuffer = Buffer.from(withAppKeyOrigin, 'utf8');
       const appKeyBuffer = Buffer.concat([privateKey, appKeyOriginBuffer]);
-      const appKeyPrivateKey = arrToBufArr(keccak256(appKeyBuffer));
-      const appKeyPublicKey = privateToPublic(appKeyPrivateKey);
+      const appKeyPrivateKey = Buffer.from(keccak256(appKeyBuffer));
```
danroc, line 564: Is it expected to still use `Buffer` instead of `Uint8Array`?

packages/keyring-eth-trezor/src/trezor-keyring.ts
```diff
@@ -542,13 +548,13 @@ export class TrezorKeyring implements Keyring {
   }
 
   #normalize(buf: Buffer): string {
-    return bufferToHex(buf).toString();
+    return bytesToHex(buf).toString();
```
danroc, line 551: Is the `.toString()` call needed?

packages/keyring-eth-hd/src/hd-keyring.ts
matthewwalsh0: I mentioned in the PR description that I wanted to do the minimum changes for the sake of stability, so some of the private and helper methods still rely on `Buffer` which I haven't changed.

It should be harmless since `Buffer` is a subset of `Uint8Array` meaning every `Buffer` is an `Uint8Array`, where as every `Uint8Array` isn't a `Buffer`.

packages/keyring-eth-hd/src/hd-keyring.ts
matthewwalsh0: No as we were previously using `bytesToHex` from `cryptography/utils` which doesn't include the `0x` prefix, so this just aligns with that but using the `@metamask/utils` version for consistency and performance.

packages/keyring-eth-trezor/src/trezor-keyring.ts
matthewwalsh0: Removed.

danroc (approved):

PR #222, opened 2025-02-19
feat(keyring-utils): add `isScopeEqual` + `isScopeEqualToAny`

ccharly:
Adding new util functions to compare 2 scopes (CAIP-2 chain ID) together (also supporting the special EOA case: `eip155:0`).

### DIFF
--- a/packages/keyring-utils/src/scopes.ts
+++ b/packages/keyring-utils/src/scopes.ts
@@ -1,0 +1,50 @@
+import type { CaipChainId } from '@metamask/utils';
+import { KnownCaipNamespace } from '@metamask/utils';
+
+// We do not uses the `EthScope` from `@metamask/keyring-api` here, to avoid pulling it
+// as a dependency.
+export const ETH_SCOPE_EOA = `${KnownCaipNamespace.Eip155}:0`;
+
+// We use a string prefix when comparing EOAs to avoid unecessary splits.
+const ETH_SCOPE_PREFIX = `${KnownCaipNamespace.Eip155}:`;
+
+/**
+ * Check if scope is compatible with a list of scopes. It also supports the special
+ * case of `eip155:0` for EVM EOA chain ID which is compatible with any EVM chain
+ * ID (`eip155:*`).
+ *
+ * @param scope - The scope (CAIP-2 chain ID) to check.
+ * @param other - Another scope to compare to.
+ * @returns True if both scope are compatible, false otherwise.
+ */
+export function isScopeEqual(scope: CaipChainId, other: CaipChainId): boolean {
+  const isScopeEoa = scope === ETH_SCOPE_EOA;
+  const isOtherEoa = other === ETH_SCOPE_EOA;
+
+  // Special case for EOA scopes (we check on both sides);
+  if (isScopeEoa) {
+    return other.startsWith(ETH_SCOPE_PREFIX);
+  }
+  if (isOtherEoa) {
+    return scope.startsWith(ETH_SCOPE_PREFIX);
+  }
+
+  // Normal case, if both scope stricly matches, then they are compatible.
+  return scope === other;
+}
+
+/**
+ * Check if scope is compatible with a list of scopes. It also supports the special
+ * case of `eip155:0` for EVM EOA chain ID which is compatible with any EVM chain
+ * ID (`eip155:*`).
+ *
+ * @param scope - The scope (CAIP-2 chain ID) to check.
+ * @param scopes - The list of scopes to check against.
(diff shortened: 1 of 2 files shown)


packages/keyring-utils/src/scopes.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @param scope - The scope (CAIP-2 chain ID) to check.
+ * @param other - Another scope to compare to.
+ * @returns True if both scope are compatible, false otherwise.
+ */
+export function isScopeEqual(scope: CaipChainId, other: CaipChainId): boolean {
+  const isScopeEoa = scope === ETH_SCOPE_EOA;
+  const isOtherEoa = other === ETH_SCOPE_EOA;
+
+  // Special case for EOA scopes (we check on both sides);
```
danroc, line 24:
```suggestion
  // Special case for EOA scopes (we check on both sides).
```

packages/keyring-utils/src/scopes.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (isScopeEoa) {
+    return other.startsWith(ETH_SCOPE_PREFIX);
+  }
+  if (isOtherEoa) {
+    return scope.startsWith(ETH_SCOPE_PREFIX);
+  }
+
+  // Normal case, if both scope stricly matches, then they are compatible.
```
danroc, line 32:
```suggestion
  // Normal case, if both scopes strictly matches, then they are compatible.
```

packages/keyring-utils/src/scopes.ts
```diff
@@ -0,0 +1,50 @@
+import type { CaipChainId } from '@metamask/utils';
+import { KnownCaipNamespace } from '@metamask/utils';
+
+// We do not uses the `EthScope` from `@metamask/keyring-api` here, to avoid pulling it
```
danroc, line 4:
```suggestion
// We do not use the `EthScope` from `@metamask/keyring-api` here, to avoid pulling it
```

packages/keyring-utils/src/scopes.ts
```diff
@@ -0,0 +1,50 @@
...
+// as a dependency.
+export const ETH_SCOPE_EOA = `${KnownCaipNamespace.Eip155}:0`;
+
+// We use a string prefix when comparing EOAs to avoid unecessary splits.
+const ETH_SCOPE_PREFIX = `${KnownCaipNamespace.Eip155}:`;
+
+/**
+ * Check if scope is compatible with a list of scopes. It also supports the special
```
danroc, line 12: Wrong copy+paste?

packages/keyring-utils/src/scopes.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (isOtherEoa) {
+    return scope.startsWith(ETH_SCOPE_PREFIX);
+  }
+
+  // Normal case, if both scope stricly matches, then they are compatible.
+  return scope === other;
+}
+
+/**
+ * Check if scope is compatible with a list of scopes. It also supports the special
```
danroc, line 37:
```suggestion
 * Check if `scope` matches any scope from `scopes`. It also supports the special
```

packages/keyring-utils/src/scopes.ts
```diff
@@ -0,0 +1,50 @@
...
+ * case of `eip155:0` for EVM EOA chain ID which is compatible with any EVM chain
+ * ID (`eip155:*`).
+ *
+ * @param scope - The scope (CAIP-2 chain ID) to check.
+ * @param scopes - The list of scopes to check against.
+ * @returns True if the scope is compatible within the list of scopes.
```
danroc, line 43:
```suggestion
 * @returns True if `scope` matches any scope from `scopes`, false otherwise.
```

packages/keyring-utils/src/scopes.ts
```diff
@@ -0,0 +1,50 @@
...
+
+// We do not use the `EthScope` from `@metamask/keyring-api` here to avoid pulling it
+// as a dependency.
+export const ETH_SCOPE_EOA = `${KnownCaipNamespace.Eip155}:0`;
+
+// We use a string prefix when comparing EOAs to avoid unecessary splits.
+const ETH_SCOPE_PREFIX = `${KnownCaipNamespace.Eip155}:`;
+
+/**
+ * Check if scope is matching with another scopes. It also supports the special
```
danroc, line 12:
```suggestion
 * Check if scope matches with another scope. It also supports the special
```

danroc (approved):

PR #241, opened 2025-03-03
chore: update `@trezor/connect-web`

ccharly:
Similar to:
- https://github.com/MetaMask/accounts/pull/234

But just use a stricter version range to match the one from the extension, see:
- https://github.com/MetaMask/metamask-extension/pull/30640

We should remove the `syncpack` rule once we agree on moving on to the next minor (or higher) of this package. For now it has been decided to only bump the patch version to avoid introducing too many changes for Trezor support.

### DIFF
--- a/.syncpackrc
+++ b/.syncpackrc
@@ -16,6 +16,11 @@
       "range": "~"
     },
     {
+      "label": "use tilde range for @trezor/connect-web (TODO: remove once the extension agree on using a higher version)",
+      "dependencies": ["@trezor/connect-web"],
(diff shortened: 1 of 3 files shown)


mikesposito (approved):

ccharly:
@SocketSecurity ignore npm/@solana/rpc-transport-http@2.0.0

We already ignored it elsewhere, see here for details: https://github.com/MetaMask/metamask-extension/pull/30640#issuecomment-2690918355

darioAnongba (approved):

PR #244, opened 2025-03-06
feat(keyring-eth-hd): store mnemonic seed in `HdKeyring` class

FrederikBolding:
Store the computed mnemonic seed in the `HdKeyring` class which allows it to be re-used by other parts of MetaMask doing key derivation on compatible curves.

This will unlock some performance improvements in the Snaps platform that currently uses the mnemonic and recomputes the seed for key derivation.

### DIFF
--- a/packages/keyring-eth-hd/src/hd-keyring.test.ts
+++ b/packages/keyring-eth-hd/src/hd-keyring.test.ts
@@ -19,9 +19,12 @@ import {
   type MessageTypes,
   type EIP7702Authorization,
 } from '@metamask/eth-sig-util';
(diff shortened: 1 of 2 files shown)


mikesposito (commented):
Gorgeous!

Only one thing: should `this.seed` be unset [here](https://github.com/MetaMask/accounts/blob/68a14b024b8f426aa014921b2897ad059054cb86/packages/keyring-eth-hd/src/hd-keyring.ts#L156-L156) along with the mnemonic?

mikesposito (approved):

danroc (approved):

PR #249, opened 2025-03-13
fix(keyring-eth-ledger-bridge): fix use of `salt` value as a `string` for `signTypedData<V4>`

dawnseeker8:
This PR is to fix the extension bug [30473](https://github.com/MetaMask/metamask-extension/issues/30473)

Basically the issue was caused by domain salt was defined as string in client DApp, and @metamask/eth-sig-util library we used define salt as arrayType. which cause miss match with previous code of salt value.

This code has been tested with client dapp and our e2e test-dapp to confirm it will work for both.


dawnseeker8:
@metamaskbot publish-preview

dawnseeker8:
@metamaskbot publish-preview

dawnseeker8:
@metamaskbot publish-preview

dawnseeker8:
@metamaskbot publish-preview

packages/keyring-eth-ledger-bridge/src/ledger-keyring.test.ts
```diff
@@ -1024,6 +1069,55 @@ describe('LedgerKeyring', function () {
         );
       });
 
+      it('resolves properly when message domain salt is an ArrayBuffer', async function () {
```
ccharly, line 1072: Since the initial bug was to also cover `salt` as a `string` (if I understood correctly), shouldn't we have a test for this as well?

packages/keyring-eth-ledger-bridge/src/ledger-keyring.test.ts
```diff
@@ -1127,4 +1221,88 @@ describe('LedgerKeyring', function () {
       });
     });
   });
+
+  describe('private methods', function () {
+    describe('#getApiUrl', function () {
```
ccharly, line 1226: Are those also related to that bug?

If not, it would be better to have them on a separate PR IMO.

dawnseeker8: no, all this added unit tests relative to passing the unit tests coverageThreshold, i dont know why this branch have quite high threshold and without those unit tests added, unit tests start to failed. that is why i added a lot of unit tests in both ledger-keyring and ledger-iframe-brdige.

packages/keyring-eth-ledger-bridge/jest.config.js
```diff
@@ -23,10 +23,10 @@ module.exports = merge(baseConfig, {
...
   coverageThreshold: {
     global: {
-      branches: 92.41,
-      functions: 97.89,
-      lines: 97.18,
-      statements: 97.21,
+      branches: 91.78,
+      functions: 97.87,
+      lines: 97.15,
+      statements: 97.18,
```
ccharly, line 29: We're still covering the same code paths, so that's ok if those threshold are getting lowered a bit.

ccharly (approved):

PR #252, opened 2025-03-18
feat!: add `SnapKeyring.createAccount` + internal options

ccharly:
Introducing a new way of handling internal options (`display*` flags) that were used to configure the account creation flow.

Initially those flags were being sent by the Snap (only interpreted for preinstalled Snaps) itself to allow to skip some specific steps of the flow (e.g display the account name input OR display the confirmation about the account creation action).

The main problem with the previous approach is that the caller (MetaMask here) of `keyring_createAccount` cannot skip those steps programmatically. It's up to the Snap to the Snap to emit the right flags with the correct values during the flow.

- - -

With the new approach, we "hide" those flags behind a context that is owned by the Snap keyring. In addition to that, we've added a new `SnapKeyring.createAccount` method that allows the caller (MetaMask here, again) to enable those flags if needed.

This way, we can configure the flow to behave differently:
```ts
const BITCOIN_SNAP_ID = 'npm:...';
const SOLANA_SNAP_ID = 'npm:...';

const keyring = new SnapKeyring(...);

// Basic flow
keyring.createAccount(SOLANA_SNAP_ID, { scope: '...' }, {
  // We don't want to display those intermediary screens.
  displayConfirmations: false,
});

// Creating a Solana account "on-the-fly" with a default name
keyring.createAccount(SOLANA_SNAP_ID, { scope: '...' }, {
  // We don't want to display those intermediary screens.
  displayConfirmations: false,
  // Will defaults to a generic account name.
  displayAccountNameSuggestion: false,
});
```

> [!CAUTION]
> The current flow will still select the newly created account at the end of the flow, but with this new approach we could easily extend our "internal options" and skip that step too, like
> ```ts
> // Could be used by the new bridge EVM <> Solana flow
> // OR could be used by profile-sync to create account without selecting them.
> keyring.createAccount(SOLANA_SNAP_ID, { scope: '...' }, {
>  displayConfirmations: false,
>  displayAccountNameSuggestion: false,
>  // The Snap keyring won't select this account automatically.
>  setSelectedAccount: false,
> });
> ```

### DIFF
--- a/packages/keyring-api/src/api/keyring.ts
+++ b/packages/keyring-api/src/api/keyring.ts
@@ -9,6 +9,7 @@ import type { ResolvedAccountAddress } from './address';
 import type { Balance } from './balance';
 import type { CaipChainId, CaipAssetType, CaipAssetTypeOrId } from './caip';
 import type { KeyringAccountData } from './export';
+import type { InternalOptions } from './options';
 import type { Paginated, Pagination } from './pagination';
 import type { KeyringRequest } from './request';
 import type { KeyringResponse } from './response';
@@ -47,11 +48,14 @@ export type Keyring = {
    *
    * Creates a new account with optional, keyring-defined, account options.
    *
-   * @param options - Keyring-defined options for the account (optional).
+   * @param options - Keyring-defined options for the account (optional). The
+   * 'metamask' internal options needs to be re-emitted during `notify:*` events.
    * @returns A promise that resolves to the newly created KeyringAccount
    * object without any private information.
    */
-  createAccount(options?: Record<string, Json>): Promise<KeyringAccount>;
+  createAccount(
+    options?: Record<string, Json> & InternalOptions,
+  ): Promise<KeyringAccount>;
 
   /**
    * Lists the assets of an account (fungibles and non-fungibles) represented
--- a/packages/keyring-api/src/api/keyring.ts
+++ b/packages/keyring-api/src/api/keyring.ts
@@ -9,7 +9,7 @@ import type { ResolvedAccountAddress } from './address';
 import type { Balance } from './balance';
 import type { CaipChainId, CaipAssetType, CaipAssetTypeOrId } from './caip';
 import type { KeyringAccountData } from './export';
-import type { InternalOptions } from './options';
+import type { MetaMaskOptions } from './options';
 import type { Paginated, Pagination } from './pagination';
 import type { KeyringRequest } from './request';
 import type { KeyringResponse } from './response';
@@ -54,7 +54,7 @@ export type Keyring = {
    * object without any private information.
    */
   createAccount(
-    options?: Record<string, Json> & InternalOptions,
+    options?: Record<string, Json> & MetaMaskOptions,
   ): Promise<KeyringAccount>;
 
   /**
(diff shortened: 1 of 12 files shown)


ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

ccharly (commented):
Review session with @danroc

ccharly:
@metamaskbot publish-preview

packages/keyring-api/src/api/keyring.ts
```diff
@@ -47,11 +48,14 @@ export type Keyring = {
    *
    * Creates a new account with optional, keyring-defined, account options.
    *
-   * @param options - Keyring-defined options for the account (optional).
+   * @param options - Keyring-defined options for the account (optional). The
+   * 'metamask' internal options needs to be re-emitted during `notify:*` events.
```
danroc, line 52: Can you remind me what do you mean by:
> The 'metamask' internal options needs to be re-emitted during `notify:*` events.

danroc: nvm, I got it from the code below :D

packages/keyring-snap-bridge/src/SnapIdMap.ts
```diff
@@ -132,6 +132,35 @@ export class SnapIdMap<Value extends { snapId: SnapId }> {
...
+   * ```
+   *
+   * @param snapId - Snap ID present in the value to get.
+   * @param key - Key of the element to get.
+   * @returns The value associated with the given key and Snap ID.
+   */
+  pop(snapId: SnapId, key: string): Value | undefined {
+    const value = this.get(snapId, key);
+
+    if (value !== undefined) {
```
danroc, line 158: Just a note: you don't need to check before calling `this.delete()`, but I think it's fine both ways.

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -195,9 +209,44 @@ export class SnapKeyring extends EventEmitter {
...
+   * @returns Internal options if found, or defaults internal options values otherwise.
+   */
+  #getInternalOptionsOrDefaults(
+    snapId: SnapId,
+    correlationId: string | undefined,
+  ): SnapKeyringInternalOptions {
+    if (correlationId) {
+      // We still need to check if the correlation ID is valid and associated to
+      // some internal options.
+      // NOTE: `found` will be `undefined` if a Snap tried to use a correlation ID that
```
danroc, line 232:
```suggestion
      // some internal options.
      //
      // NOTE: `found` will be `undefined` if a Snap tried to use a correlation ID that
```

packages/keyring-snap-bridge/src/SnapKeyring.test.ts
```diff
@@ -427,11 +451,13 @@ describe('SnapKeyring', () => {
...
 
             const params = {
-              account: account as unknown as KeyringAccount,
-              ...(displayConfirmation !== undefined && { displayConfirmation }),
+              account,
               ...(accountNameSuggestion !== undefined && {
                 accountNameSuggestion,
               }),
+              // Those flags have now been deprecated, but some older Snaps might still use them, so we
+              // must be able to use them in the `AccountCreated` without throwing any error.
```
danroc, line 459: Are they used if set?

ccharly: Nope, not anymore. I was wondering if that was useful or not...

packages/keyring-snap-bridge/src/SnapIdMap.ts
ccharly: Ah yes true 🙈 well I might omit the `if` then, it'd be a bit more "concise" that way.

danroc (approved):
LGTM

PR #253, opened 2025-03-19
fix(keyring-eth-ledger-bridge): trim `0x` prefix for  `signTransaction` (due to latest `@ethereumjs` upgrade)

dawnseeker8:
This PR will fix the ledger sign transaction issue due to this ethereumjs upgrade PR https://github.com/MetaMask/accounts/pull/209 

and basically the ethereumjs upgrade code change make the transaction rawTxHex contain `0x` prefix which is not valid for to @ledgerhq/hw-eth-app clearSignTransaction function. that functions expect a rawTxHex without `0x` prefix.


ccharly (approved):
Code LGTM, I haven't tested it myself though.

PR #259, opened 2025-03-21
feat(keyring-snap-bridge): add TTL for internal options

ccharly:
Adding a TTL to the new internal options features to avoid having leaking resources if a Snap does not properly re-emit its correlation ID.

### DIFF
--- a/packages/keyring-snap-bridge/src/SnapKeyring.ts
+++ b/packages/keyring-snap-bridge/src/SnapKeyring.ts
@@ -59,6 +59,7 @@ import {
 import { projectLogger as log } from './logger';
 import { isAccountV1, migrateAccountV1 } from './migrations';
 import {
+  DEFAULT_INTERNAL_OPTIONS_TTL_SECS,
   getDefaultInternalOptions,
   type SnapKeyringInternalOptions,
 } from './options';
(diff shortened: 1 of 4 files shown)


packages/keyring-snap-bridge/src/options.ts
```diff
@@ -1,3 +1,9 @@
+/**
+ * Default TTL for internal options. Past that time, the internal options will
+ * be considered outdated and will be removed automatically.
+ */
+export const DEFAULT_INTERNAL_OPTIONS_TTL_SECS = 60 * 60; // 1 hour
```
gantunesr, line 5: Is there any motivation for it to be one hour or was it chosen arbitrarily?

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -248,6 +258,31 @@ export class SnapKeyring extends EventEmitter {
     return getDefaultInternalOptions();
   }
 
+  /**
+   * Cleanup outdated internal options.
+   *
+   * @returns The number of entries that have been deleted.
```
gantunesr, line 264: Why do you need this return?

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -347,6 +382,10 @@ export class SnapKeyring extends EventEmitter {
       this.#getInternalOptionsOrDefaults(snapId, metamask?.correlationId),
     );
 
+    // To avoid having a `setInterval` to do the cleanup, we run it explicitly from
+    // time to time.
+    await this.cleanupInternalOptions();
```
gantunesr, line 387: Will this check all snaps for each account creation event?

PR #261, opened 2025-03-21
refactor(keyring-snap-bridge): re-allow Snap to some internal options

ccharly:
Re-allow a Snap to send `display*` flags during the `notify:accountCreated` event.

### DIFF
--- a/packages/keyring-snap-bridge/src/SnapKeyring.ts
+++ b/packages/keyring-snap-bridge/src/SnapKeyring.ts
@@ -59,7 +59,7 @@ import {
 import { projectLogger as log } from './logger';
 import { isAccountV1, migrateAccountV1 } from './migrations';
 import {
-  getDefaultInternalOptions,
+  getInternalOptionsOf,
   type SnapKeyringInternalOptions,
 } from './options';
 import { SnapIdMap } from './SnapIdMap';
@@ -222,10 +222,10 @@ export class SnapKeyring extends EventEmitter {
    * @param correlationId - Correlation ID associated with the internal options.
    * @returns Internal options if found, or defaults internal options values otherwise.
    */
-  #getInternalOptionsOrDefaults(
+  #getInternalOptions(
     snapId: SnapId,
     correlationId: string | undefined,
-  ): SnapKeyringInternalOptions {
+  ): SnapKeyringInternalOptions | undefined {
     if (correlationId) {
       // We still need to check if the correlation ID is valid and associated to
       // some internal options.
@@ -245,7 +245,7 @@ export class SnapKeyring extends EventEmitter {
       );
     }
 
-    return getDefaultInternalOptions();
+    return undefined;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -222,10 +222,10 @@ export class SnapKeyring extends EventEmitter {
    * @param correlationId - Correlation ID associated with the internal options.
    * @returns Internal options if found, or defaults internal options values otherwise.
```
danroc, line 223: Update `returns`

packages/keyring-snap-bridge/src/options.ts
```diff
@@ -36,3 +36,47 @@ export function getDefaultInternalOptions(): Required<SnapKeyringInternalOptions
...
+};
+
+/**
+ * Combine multiple internal options together. The internal options
+ * in the list will be merged in the order they are provided. If some options
+ * are still not defined, it will fallback to the default ones.
+ *
+ * @param internalOptions - List of internal options to combine.
+ * @returns Computed internal options.
+ */
```
danroc, line 54:
```suggestion
/**
 * Combine multiple internal options together. The internal options
 * in the list will be merged in the order they are provided. The first
 * appearance of a key is used, and the following are ignored. If some options
 * are still not defined, it will fallback to the default ones.
 *
 * @param internalOptions - List of internal options to combine.
 * @returns Computed internal options.
 */
```

packages/keyring-snap-bridge/src/options.ts
```diff
@@ -36,3 +36,47 @@ export function getDefaultInternalOptions(): Required<SnapKeyringInternalOptions
...
+
+  const defaults: Required<SnapKeyringInternalOptions> =
+    getDefaultInternalOptions();
+
+  const keys = Object.keys(defaults) as (keyof SnapKeyringInternalOptions)[];
+  for (const key of keys) {
+    // We start of with `undefined` and check every options of the list, if
+    // we find any that is defined, we used it. In the end, we will fallback
+    // to the default ones that are guaranteed to be defined (thanks to the
+    // use of `Required`).
```
danroc, line 68:
```suggestion
    // We start off with `undefined` and check every option in the list. If
    // we find any that is defined, we use it. In the end, we will fallback
    // to the default ones that are guaranteed to be defined (thanks to the
    // use of `Required`).
```

packages/keyring-snap-bridge/src/options.ts
danroc, line 54:
```suggestion
/**
 * Combine multiple internal options together. The internal options
 * in the list will be merged in the order they are provided. The first
 * appearance of a key is used, and the following are ignored. If some options
 * are still not defined, it will fallback to the default ones.
 *
 * @param internalOptions - List of internal options to combine.
 * @returns Computed internal options.
 */
```

PR #266, opened 2025-03-28
feat: add discoverAccounts keyring method

ccharly:
This new method will allow a MetaMask clients to discover potential accounts that have an history on the blockchain (e.g when importing a new SRP to MetaMask).

This method won't create the accounts directly, it's the client itself that will decide to create them or not.

Since discovering accounts might requires some "computation", a Snap could decide to "cache" the information/data he discovered for all those accounts, so they can be re-used later on once the account is created.


ccharly:
@metamaskbot publish-preview

packages/keyring-api/src/api/keyring.ts
```diff
@@ -84,6 +86,20 @@ export type Keyring = {
...
   ): Promise<Paginated<Transaction>>;
 
+  /**
+   * Discover accounts.
+   *
+   * @param scopes - List of scopes to discover.
+   * @param entropySource - The entropy source being used to derive accounts.
+   * @param groupeIndex - The group index for this account (could be used as the `account_index` in BIP-44).
+   * @returns A promise that resolves to a list of discovered accounts.
+   */
```
danroc, line 96:
```suggestion
  /**
   * Discovers accounts.  
   *  
   * The `groupIndex` is used to group accounts with the same value.  
   * In strictly BIP-44 wallets, it matches `account_index`, but in wallets  
   * that deviate from BIP-44 recommendations, it may align with a different  
   * path level for compatibility.  
   *  
   * @param scopes - The list of scopes for account discovery.  
   * @param entropySource - The entropy source used for deriving accounts.  
   * @param groupIndex - The group index for this account.  
   * @returns A promise resolving to a list of discovered accounts.  
   */ 
```

packages/keyring-api/src/api/keyring.ts
```diff
@@ -84,6 +86,25 @@ export type Keyring = {
...
+   * The `groupIndex` is used to group accounts with the same value.  
+   * In strictly BIP-44 wallets, it matches `account_index`, but in wallets  
+   * that deviate from BIP-44 recommendations, it may align with a different  
+   * path level for compatibility.  
+   *  
+   * @param scopes - The list of scopes for account discovery.  
+   * @param entropySource - The entropy source used for deriving accounts.  
+   * @param groupIndex - The group index for this account.  
+   * @returns A promise resolving to a list of discovered accounts.  
+   */
```
danroc, line 101:
```suggestion
  /**
   * Discover accounts.
   *
   * This method is called by the client to allow the keyring to discover
   * existing accounts based on the provided scopes and entropy source ID. It's
   * considered existing accounts, accounts that have at least one transaction,
   * as per BIP-44.
   *
   * The `groupIndex` is used to group accounts with the same value. In
   * strictly BIP-44 wallets, it matches `account_index`, but in wallets that
   * deviate from BIP-44 recommendations, it may align with a different path
   * level for compatibility.
   *
   * @param scopes - The list of scopes for account discovery.
   * @param entropySource - The ID of the entropy source used to derive the accounts.
   * @param groupIndex - The group index that should be used to derive the accounts.
   * @returns A promise resolving to a list of discovered accounts.
   */
```

danroc (approved):

PR #273, opened 2025-04-02
feat!: add `KeyringRequest.origin`

ccharly:
Adding a `origin` field to `KeyringRequest`. This will be used to provide the actual caller's origin (dapp origin in this case) to the Snap, so we can use it/display it when we need strong confirmations from a user.

The normal request `origin` is not enough since `submitRequest` will be actually dispatched by MM itself, but we still keep track of the initial caller in some other context, and this one got re-injected directly on the "keyring request".


packages/keyring-snap-client/src/RestrictedKeyringClient.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Extract the method name from a RPC method name.
+ */
+type KeyringRpcMethodToMethodName<RpcMethod extends string> =
+  RpcMethod extends `keyring_${infer Method}` ? Method : never;
+
+/**
+ * Restricted methods list.
+ */
+export const RestrictedKeyringRpcMethod = [
```
danroc, line 19: I think **restricted** is not the best name here, it gives the impression that it's something that requires more rights, which is not completely correct. Maybe `Exposed*`, `External*`, `Public*`, or even `CompanionDapp*`.

ccharly: Yes, that's why I wanted your opinion on this 😄

I also thought of `Public*` TBH, but `CompanionDapp*` is also a good candidate 🤔 basically I wanted to expose that somewhere on the `accounts`-packages, so we can re-use it afterward on the extension/mobile-app.

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -1123,6 +1135,7 @@ export class SnapKeyring extends EventEmitter {
     });
 
     const signedTx = await this.#submitRequest({
+      origin: 'metamask',
```
danroc, line 1138: > **Note:** we are setting the origin to `metamask` here regardless of the real origin.

Do you have an idea how we are going to send the real origin? Also, will this new field break the existing Snaps?

ccharly: Good point 🤔 I wrongly assumed we were using those methods when interacting through the MetaMask UI, but this will also be called externally with call like `eth_sign*`, in which case, the true `origin` would not be `metamask`.

And yes, this new field might also break existing Snaps. I don't know else I could trick that. My only idea right now would be to "version the API", and have a runtime check on the version, something like:

```ts
getVersion() {
  const { version } = await this.#snapClient
    .withSnapId(snapId)
    .getVersion();
    
  if (!version) {
    return ...; // First version when we introduced the new `getVersion`?
  }
}
```

But even know, we cannot check which methods are being implemented on a Snap, which would result in a runtime error for this ⬆️.

And adding a version won't solve everything, cause we would have type-issue here to (here we need to pass `origin`, or we need to have another variant type for `submitRequest` with an optional `origin`, that won't scale well IMO..)

ccharly:
@metamaskbot publish-preview

package.json
```diff
@@ -34,6 +34,7 @@
     "test:types": "yarn foreach test:types"
   },
   "resolutions": {
+    "@metamask/snaps-sdk@^6.24.0": "npm:@metamask/snaps-sdk@^7.0.0",
```
ccharly, line 37: This one is required because the latest `@metamask/snaps-utils` is using this version and not the `@metamask/snaps-sdk@^7.0.0` version.

An alternative would be to not use the `@metamask/snaps-sdk@^7.0.0` ourselves, but I thought it made sense to use it here, given that this change is "associated" with this new platform version (=== Snaps SDK version).

ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

danroc:
I reviewed this with @ccharly on a call. LGTM. Key points:

- We currently send `metamask` as the origin for EVM signing requests. This is problematic but requires significant client changes to fix.

- ✅ `PLATFORM_VERSION_TO_KEYRING_VERSION` was previously a plain object, which caused issues due to order sensitivity. It's now an array to preserve order.

- ✅ The `send` method in the internal keyring client was public. This was resolved by extending `KeyringInternalSnapClient` with a `submitRequestV1` method.

- The `@metamask/keyring-api` dependency in `keyring-snap-sdk` should likely be a `peerDependency`. This will be addressed in a separate PR.

packages/keyring-snap-bridge/src/SnapKeyring.ts
ccharly: Just for the record, we solved this problem thanks to the new `SnapController:isMinimumPlatformVersion` action + introducing a new platform version (7.0.0) which would act as the next version to introduce this breaking change.

Snap that declares this version (and higher) are expected to use the new `KeyringRequest.origin`.

Older Snaps (<7.0.0) won't receive this new field to keep the retro-compatibility.

PR #279, opened 2025-04-14
refactor: remove timeout for messages requiring user actions

mikesposito:
https://github.com/MetaMask/accounts/pull/271 introduced timeouts for all messages between the Ledger bridge and the iframe. Though a 20-second timeout may not be enough for the user to grab the device and accept the request - especially if the user wants to look closely at the signature request on the hardware device.

For this reason, the timeout is being removed for messages requiring user intervention, while keeping it for messages that do not expect user actions (i.e. `updateTransportMethod` and `attemptMakeApp`)

### DIFF
--- a/packages/keyring-eth-ledger-bridge/src/ledger-iframe-bridge.ts
+++ b/packages/keyring-eth-ledger-bridge/src/ledger-iframe-bridge.ts
@@ -16,8 +16,6 @@ const LEDGER_IFRAME_ID = 'LEDGER-IFRAME';
 
 const IFRAME_MESSAGE_TIMEOUT = 4000;
 
-const USER_ACTION_REQUEST_TIMEOUT = 20_000;
-
 export enum IFrameMessageAction {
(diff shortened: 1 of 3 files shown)


packages/keyring-eth-ledger-bridge/src/ledger-iframe-bridge.ts
```diff
@@ -338,7 +319,7 @@ export class LedgerIframeBridge
 
   async #sendMessage<TAction extends IFrameMessageAction>(
     message: IFrameMessage<TAction>,
-    timeout = IFRAME_MESSAGE_TIMEOUT,
+    { timeout }: { timeout?: number } = {},
```
mikesposito, line 322: Matching the signature of the `#sendMessage` method we are adding on the MV3 extension offscreen iframe bridge: https://github.com/MetaMask/metamask-extension/pull/31766

gantunesr (approved):

Gudahtt (approved):
LGTM!

PR #282, opened 2025-04-30
refactor: remove account.metadata.snap.{name,enabled}

ccharly:
Those metadata should be pulled out from the `SnapController` itself to avoid duplicated state and potential inconsistencies.

This would also remove some non-needed state-watch on the `AccountsController` (since we are listening to `SnapController:stateChange` to update those accordingly, thus triggering some other `AccountsController:stateChange` which could be avoided).


ccharly:
@metamaskbot publish-preview

PR #284, opened 2025-05-05
feat: bitcoin account types

darioAnongba:
- Adds support for remaining Bitcoin account types.
- Adds stricter validation checks for the decoding of the addresses. Removes `bech32` dependency in favor of `bitcoin-address-validation`. 
- Removes size limit of scopes for Bitcoin accounts as testnet accounts span multiple scopes. `testnet`, `testnet4` and `signet` are the same account (same address).

`p2sh` is a shortcut for `p2wpkh-p2sh` (nested segwit).

* Fixes: https://consensyssoftware.atlassian.net/browse/NNT-329

### DIFF
--- a/packages/keyring-api/src/btc/types.test.ts
+++ b/packages/keyring-api/src/btc/types.test.ts
@@ -14,6 +14,18 @@ import {
 } from './types';
 import { BtcAccountType } from '../api';
 
+const BTC_P2PKH_MAINNET_ADDRESS = '1AXaVdPBb6zqrTMb6ebrBb9g3JmeAPGeCF';
+const BTC_P2SH_MAINNET_ADDRESS = '3KQPirCGGbVyWJLGuWN6VPC7uLeiarYB7x';
+const BTC_P2WPKH_MAINNET_ADDRESS = 'bc1q4degm5k044n9xv3ds7d8l6hfavydte6wn6sesw';
+const BTC_P2TR_MAINNET_ADDRESS =
+  'bc1pxfxst7zrkw39vzh0pchq5ey0q7z6u739cudhz5vmg89wa4kyyp9qzrf5sp';
+
+const BTC_P2PKH_TESTNET_ADDRESS = 'mrDHfcAPosFsabxBKe2U3EdxX5Kph8Zd4f';
+const BTC_P2SH_TESTNET_ADDRESS = '2N7AeKCw7p8uRRQjXPeHW7UPGhR8LYHEzBT';
+const BTC_P2WPKH_TESTNET_ADDRESS = 'tb1qqecaw32rvyjgez706t5chpr8gan49wfuk94t3g';
+const BTC_P2TR_TESTNET_ADDRESS =
+  'tb1p6epn3ctassfp54lnztshnpfjekn7khyarrnm6f0yv738lgc53xxsgevs8k';
+
 describe('types', () => {
   const mockAccount = {
     id: '55583f38-d81b-48f8-8494-fc543c2b5c95',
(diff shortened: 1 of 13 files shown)


packages/keyring-api/src/btc/types.test.ts
```diff
@@ -17,49 +13,13 @@ const MOCK_ACCOUNT = {
...
-    });
-
-    it('throws an error if there no seperator', () => {
-      const address = 'bc0qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4';
-      expect(() => BtcP2wpkhAddressStruct.assert(address)).toThrow(
-        `${errorPrefix}: No separator character for ${address}`,
-      );
-    });
-
+  describe('BtcP2wpkhAccountStruct', () => {
```
ccharly, line 16:
```suggestion
  describe('BtcAccountStruct', () => {
```

packages/keyring-api/src/btc/types.ts
```diff
@@ -39,18 +16,18 @@ export enum BtcMethod {
...
 
-export const BtcP2wpkhAccountStruct = object({
+export const BtcAccountStruct = object({
   ...KeyringAccountStruct.schema,
 
   /**
    * Account address.
    */
-  address: BtcP2wpkhAddressStruct,
+  address: string(),
```
ccharly, line 25: I think we should still validate the address formats here. We could use `bitcoin-address-validation` package for this, however, given that we are using 1 single struct, it will be hard to make sure the `address` is valid for the given `type`.

So maybe we should have 4 different structs instead (similarly to `BtcP2wpkhAccountStruct`, we would have `BtcP2shAccountStruct`, etc...), allowing us to strictly check the `address` is valid for the account's `type`?

WDYT?

darioAnongba: The address should not be validated because for the extension, the address is a `string`. For example if the Snap tells you that your address is `fooBar` then that's your address. you can't decide that you don't agree with the Snap.
Knowing the address format is not needed in the business logic.

darioAnongba: After discussing, all types were added.

packages/keyring-api/src/btc/types.ts
```diff
@@ -39,30 +69,77 @@ export enum BtcMethod {
...
+   * Account type.
+   */
+  type: literal(`${BtcAccountType.P2pkh}`),
+});
+
+export const BtcP2shAccountStruct = object({
+  ...BtcAccountStruct.schema,
+
+  /**
+   * Account P2PSH address.
```
ccharly, line 104:
```suggestion
   * Account P2SH address.
```

packages/keyring-api/src/btc/types.ts
```diff
@@ -39,30 +69,77 @@ export enum BtcMethod {
...
 
-export const BtcP2wpkhAccountStruct = object({
+const BtcAccountStruct = object({
   ...KeyringAccountStruct.schema,
 
   /**
-   * Account address.
+   * Account supported scopes (CAIP-2 chain ID).
+   */
+  scopes: nonempty(array(enums(Object.values(BtcScope)))),
```
ccharly, line 78: Maybe we should keep using normal `CaipChainId` here too? I was thinking of regtests, but I know we already have a "fake" scope for this one (`bip122:regtest`) so maybe that's ok... 

For other kind of accounts, we are just using a `CaipChainId[]`, but that's mainly cause there could be new chains in the future, Bitcoin might be different here 🤔

```suggestion
  scopes: nonempty(array(CaipChainId)),
```

So maybe, for now we could just go with the "general case", WDYT?

packages/keyring-utils/src/btc/address.ts
```diff
@@ -1,4 +1,17 @@
...
+import { validate, Network, getAddressInfo } from 'bitcoin-address-validation';
+
+/**
+ * Returns whether an address is on the Bitcoin mainnet.
+ *
+ * @param address - The address to check.
+ * @param type - The address type to check.
+ * @returns `true` if the address is valid, `false` otherwise.
+ */
+export function isBtcAddress(address: string, type: AddressType): boolean {
```
ccharly, line 11: Since we're "leaking" the `AddressType` from `bitcoin-address-validation`, I wonder if we should have this function in `packages/keyring-api/src/btc/types.ts` directly, and this one would just become a `validate(address)` (without knowledge of the address's type)?

an other alternative would be to use `type: 'p2sh' | 'p2wkph' | ...` and cast it to `AddressType` (this could be slightly unsafe though..)

darioAnongba: true, I'll move it to the `types.ts` as a private function. That's what I initially did and then saw that similar functions were exported from keyring-utils.

packages/keyring-api/src/btc/types.ts
darioAnongba: the `regtest` scope doesn't change, we use `bip122:regtest` because anyone is free to change the genesis block hash but in practice it will never change: https://github.com/bitcoin/bitcoin/blob/v27.0/src/kernel/chainparams.cpp#L476. We can use the "real" scope if you prefer.

We apply the maximum level of validation to addresses, so there is no reason not to do the same for scopes because it's the same argument, we either validate the actual values (which I think is wrong) or we validate the format.
The probability that we want to support new addresses (like `p2ms` or `p2wsh`) is actually higher than the probability of supporting new scopes.

darioAnongba: moving back to `CaipChainId `

ccharly (approved):

PR #294, opened 2025-06-09
feat(keyring-eth-ledger-bridge): enhance error handling for Ledger signing operations

dawnseeker8:
Added static method to identify Ledger-specific errors and improved error handling in the LedgerKeyring class. Now, specific error messages are thrown for user-rejected transactions and when blind signing is disabled, providing clearer feedback to users during transaction signing.

### DIFF
--- a/packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
+++ b/packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
@@ -94,6 +94,17 @@ export class LedgerKeyring implements Keyring {
 
   deviceId = '';
 
+  static #isLedgerError(
+    error: unknown,
+  ): error is { errorCode: string; message: string } {
+    return (
+      typeof error === 'object' &&
+      error !== null &&
+      'errorCode' in error &&
+      'message' in error
+    );
+  }
+
   readonly type: string = keyringType;
 
   page = 0;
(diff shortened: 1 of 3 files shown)


dawnseeker8:
@metamaskbot publish-preview

dawnseeker8:
@metamaskbot publish-preview

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -413,7 +414,26 @@ export class LedgerKeyring implements Keyring {
...
+          throw new Error('Ledger: User rejected the transaction');
+        }
+
+        if (
+          transportError.statusCode === 27264 &&
+          transportError.message.includes('Invalid data received (0x6a80)')
+        ) {
+          throw new Error('Ledger: Blind signing must be enabled');
+        }
+      }
```
ccharly, line 435: Let's make a function for this, something like `reThrowIfTransportError(error)` maybe?

So we can use it here and in the other `catch` clauses.

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -448,7 +468,20 @@ export class LedgerKeyring implements Keyring {
...
+        /**
+         * for user rejected the transaction error
+         */
+        if (
+          transportError.statusCode === 27013 &&
+          transportError.message.includes('(denied by the user?) (0x6985)')
+        ) {
+          throw new Error('Ledger: User rejected the transaction');
+        }
+      }
```
ccharly, line 483: Then we can use it like:
```suggestion
      reThrowIfTransportError(error);
```

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -540,7 +573,24 @@ export class LedgerKeyring implements Keyring {
...
+          throw new Error('Ledger: User rejected the transaction');
+        }
+
+        if (
+          transportError.statusCode === 27264 &&
+          transportError.message.includes('Invalid data received (0x6a80)')
+        ) {
+          throw new Error('Ledger: Blind signing must be enabled');
+        }
+      }
```
ccharly, line 593: Then we can use it like:
```suggestion
      reThrowIfTransportError(error);
      
```

Yes there's a missing empty line here too (compared to your other 2 other `catch` clauses)

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
ccharly: Will be addressed in a follow-up PR.

PR #301, opened 2025-06-26
feat(keyring-eth-ledger-bridge): improve error handling around `TransportStatusError`

dawnseeker8:
Added a new error handling function to manage Ledger transport errors, providing user-friendly messages for common issues. This refactor improves error handling in the LedgerKeyring class by utilizing the new function, ensuring consistent error messaging across different scenarios. Additionally, updated tests to validate the new error handling logic.

### DIFF
--- a/packages/keyring-eth-ledger-bridge/src/ledger-error-handler.test.ts
+++ b/packages/keyring-eth-ledger-bridge/src/ledger-error-handler.test.ts
@@ -103,6 +103,22 @@ describe('handleLedgerTransportError', () => {
   });
 
   describe('when error is not TransportStatusError', () => {
+    it.each([
+      { tc: 'null', value: null },
+      { tc: 'undefined', value: undefined },
+      { tc: 'non-Error instances', value: 'string error' },
+      {
+        tc: 'objects without Error prototype',
+        value: { message: 'not an error' },
+      },
+    ])('creates new Error with fallback message for: $tc', ({ value }) => {
+      const throwingFunction = (): never =>
+        handleLedgerTransportError(value, fallbackMessage);
+
+      expect(throwingFunction).toThrow(Error);
+      expect(throwingFunction).toThrow(fallbackMessage);
+    });
+
     it('re-throws Error instances as-is', () => {
       const error = new Error('Original error message');
 
@@ -113,42 +129,6 @@ describe('handleLedgerTransportError', () => {
       expect(() => handleLedgerTransportError(error, fallbackMessage)).toThrow(
         error.message,
       );
-    });
-
-    it('creates new Error with fallback message for non-Error instances', () => {
-      const nonErrorValue = 'string error';
-
-      const throwingFunction = (): never =>
-        handleLedgerTransportError(nonErrorValue, fallbackMessage);
-
-      expect(throwingFunction).toThrow(Error);
-      expect(throwingFunction).toThrow(fallbackMessage);
-    });
-
-    it('creates new Error with fallback message for null', () => {
-      const throwingFunctionNull = (): never =>
-        handleLedgerTransportError(null, fallbackMessage);
-
-      expect(throwingFunctionNull).toThrow(Error);
-      expect(throwingFunctionNull).toThrow(fallbackMessage);
-    });
-
-    it('creates new Error with fallback message for undefined', () => {
-      const throwingFunctionUndefined = (): never =>
-        handleLedgerTransportError(undefined, fallbackMessage);
-
-      expect(throwingFunctionUndefined).toThrow(Error);
-      expect(throwingFunctionUndefined).toThrow(fallbackMessage);
-    });
-
-    it('creates new Error with fallback message for objects without Error prototype', () => {
-      const plainObject = { message: 'not an error' };
-
-      const throwingFunction = (): never =>
-        handleLedgerTransportError(plainObject, fallbackMessage);
-
-      expect(throwingFunction).toThrow(Error);
-      expect(throwingFunction).toThrow(fallbackMessage);
     });
   });
 


.gitignore
```diff
@@ -75,3 +75,6 @@ node_modules/
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+
+# Cursor rules
```
dawnseeker8, line 79: this will prevent my local cursor rules to commit to project. my local cursor rules has some prompt to teach cursor model to use some default settings to better perform the tasks.

packages/keyring-eth-ledger-bridge/src/ledger-error-handler.ts
```diff
@@ -0,0 +1,52 @@
...
+          transportError.statusCode,
+          'Ledger: Ethereum app closed. Open it to unlock.',
+        );
+      default:
+        // If the status code is not one of the known codes, throw the error with the status code and message
+        throw new LedgerStatusError(
+          transportError.statusCode,
+          transportError.message,
+        );
+    }
```
ccharly, line 47: Nit: To avoid repetition here we could have introduced a function just to get the error message like:

```ts
  switch (transportError.statusCode) {
      case 0x6985:
        return 'Ledger: User rejected the transaction',
      case 0x6a80:
        return 'Ledger: Blind signing must be enabled',
      case 0x5515:
        return 'Ledger: Device is locked. Unlock it to continue.',
      case 0x650f:
        return 'Ledger: Ethereum app closed. Open it to unlock.',
      default:
        // If the status code is not one of the known codes, just use the existing error message.
        return transportError.message;
   }
```

This way, we could have used it like:
```suggestion
        throw new LedgerStatusError(
          transportError.statusCode,
          getTransportErrorMessageFrom(transportError),
        );
```

packages/keyring-eth-ledger-bridge/src/ledger-error-handler.ts
```diff
@@ -0,0 +1,52 @@
...
+    switch (transportError.statusCode) {
+      case 0x6985:
+        throw new LedgerStatusError(
+          transportError.statusCode,
+          'Ledger: User rejected the transaction',
+        );
+      case 0x6a80:
+        throw new LedgerStatusError(
+          transportError.statusCode,
+          'Ledger: Blind signing must be enabled',
```
ccharly, line 29: Same:

```suggestion
          'Ledger: Blind signing must be enabled.',
```

packages/keyring-eth-ledger-bridge/src/ledger-keyring.test.ts
```diff
@@ -492,10 +494,73 @@ describe('LedgerKeyring', function () {
...
+            result: [{ hash: '0x123' }],
+          }),
+        });
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-expect-error - Mocking window.fetch
+        global.window = { fetch: mockFetch };
+
+        await keyring.getFirstPage();
+        expect(mockFetch).toHaveBeenCalledWith(
+          expect.stringContaining('https://api-ropsten.etherscan.io'),
```
ccharly, line 563: Those tests feel completely unrelated to the transport status error topic, is that intended to have them here?

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -237,10 +237,11 @@ export class LedgerKeyring implements Keyring {
...
         hdPath: path,
       });
-    } catch (error) {
-      throw error instanceof Error
-        ? error
-        : new Error('Ledger Ethereum app closed. Open it to unlock.');
+    } catch (error: unknown) {
+      handleLedgerTransportError(
+        error,
+        'Ledger: Unknown error while unlocking account',
```
ccharly, line 243: Similar comment for all those errors, do we want to add a `.` in the end?

packages/keyring-eth-ledger-bridge/src/ledger-keyring.test.ts
```diff
@@ -1206,6 +1271,40 @@ describe('LedgerKeyring', function () {
         );
       });
 
+      it('resolves properly when message domain salt is null', async function () {
```
ccharly, line 1274: Same here.

packages/keyring-eth-ledger-bridge/src/ledger-error-handler.test.ts
```diff
@@ -0,0 +1,200 @@
...
+      statusCode,
+      message,
+      name: 'TransportStatusError',
+    };
+    Object.setPrototypeOf(error, TransportStatusError.prototype);
+    return error as TransportStatusError;
+  }
+
+  describe('when error is TransportStatusError', () => {
+    it('should handle status code 0x6985 (user rejection)', () => {
```
ccharly, line 30: We normally don't use `should ` anymore, but we prefer the imperative forms for all TC now, could you update that as well?

packages/keyring-eth-ledger-bridge/src/ledger-error-handler.test.ts
```diff
@@ -0,0 +1,200 @@
...
+    it('should re-throw Error instances as-is', () => {
+      const originalError = new Error('Original error message');
+
+      expect(() =>
+        handleLedgerTransportError(originalError, fallbackMessage),
+      ).toThrow(originalError);
+
+      expect(() =>
+        handleLedgerTransportError(originalError, fallbackMessage),
+      ).toThrow('Original error message');
```
ccharly, line 149: I guess we could just re-use the underlying message here.

```suggestion
      ).toThrow(error.message);
```

packages/keyring-eth-ledger-bridge/src/ledger-error-handler.test.ts
```diff
@@ -0,0 +1,200 @@
...
+  describe('return type', () => {
+    it('should have never return type (always throws)', () => {
+      const error = new Error('test');
+
+      // This test verifies that the function always throws
+      // by ensuring it never returns normally
+      expect(() => handleLedgerTransportError(error, fallbackMessage)).toThrow(
+        'test',
+      );
+    });
```
ccharly, line 198: Looks very similar to the other test with `originalError`.

We could do it that way (but that's gonna be a compile time error in case the signature changes):

```suggestion
    it('should have never return type (always throws)', () => {
      type ReturnTypeIsNever<Function> = ReturnType<Function> extends never ? true : false;
      
      const isNever: ReturnTypeIsNever<typeof handleLedgerTransportError> = true;
      expect(isNever).toBe(true);
```

WDYT?

Other alternative would be to use `tsd` for testing type (we do this in some other packages in that monorepo)

packages/keyring-eth-ledger-bridge/src/ledger-error-handler.test.ts
```diff
@@ -0,0 +1,200 @@
...
+      expect(() => handleLedgerTransportError(error, fallbackMessage)).toThrow(
+        LedgerStatusError,
+      );
+
+      let thrownError: unknown;
+      try {
+        handleLedgerTransportError(error, fallbackMessage);
+      } catch (error_: unknown) {
+        thrownError = error_;
+      }
```
ccharly, line 42: Could be put in a simple helper function to avoid repetition on all TC.

Alternative would be to use `it.each` otherwise.

packages/keyring-eth-ledger-bridge/src/ledger-keyring.test.ts
dawnseeker8: The tests added here is to pass the code coverage failures in CI pipeline, after i remove the methods, the code coverage drop. after analysis the code uncoverage, i decide to improve the ledger-keyring.test.ts to cover more scanario to make coverage as high as possible.

packages/keyring-eth-ledger-bridge/src/ledger-keyring.test.ts
dawnseeker8: same as above comments, also due to etheumjs/tx upgrade, the domain salt become ArrayBuffer, changing the variable name here to make us understand what type it is, even we define it as any.

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
dawnseeker8: if you look at the whole file and most error doesnt have . at the end, i think the previous error with . is wrong before. and i have changed the error message here is because `handleLedgerTransportError` functions have handled the `Ledger Ethereum app closed, Open it to unlock` situation.

packages/keyring-eth-ledger-bridge/src/ledger-error-handler.test.ts
dawnseeker8: yea, this is very good point. i have created a single helper methods called `expectLedgerStatusError` to remove the repetitive code.

packages/keyring-eth-ledger-bridge/src/ledger-keyring.test.ts
dawnseeker8: Remove unrelative tests.

packages/keyring-eth-ledger-bridge/src/ledger-keyring.test.ts
dawnseeker8: Remove unrelative tests

packages/keyring-eth-ledger-bridge/src/ledger-error-handler.test.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  describe('when error is TransportStatusError', () => {
+    it('handles status code 0x6985 (user rejection)', () => {
+      const error = createTransportStatusError('User rejected', 0x6985);
+      expectLedgerStatusError(
+        error,
+        0x6985,
+        'Ledger: User rejected the transaction',
+      );
```
ccharly, line 66: Since there's a lint warning for each tests (because there's no `expect` call in the `it` callback. WDYT about making `expectLedgerStatusError` return `true` and use it like:


```suggestion
      expect(expectLedgerStatusError(
        error,
        0x6985,
        'Ledger: User rejected the transaction',
      )).toBe(true);
```

This is super useful, but it would at least avoid the warnings everywhere.

Other options would be to use `it.each` and combine all the tests together, so we would not need the `expectLedgerStatusError`.

Something like:
```ts
it.each([
  { tc: 'user rejection', status: 0x6985, message: 'Ledger: User rejected the transaction' },
  ...
]('handles status code $status ($tc)', ({ status, error } => {
  ...
});
```

packages/keyring-eth-ledger-bridge/src/ledger-error-handler.test.ts
```diff
@@ -0,0 +1,170 @@
...
+    it('creates new Error with fallback message for objects without Error prototype', () => {
+      const plainObject = { message: 'not an error' };
+
+      const throwingFunction = (): never =>
+        handleLedgerTransportError(plainObject, fallbackMessage);
+
+      expect(throwingFunction).toThrow(Error);
+      expect(throwingFunction).toThrow(fallbackMessage);
+    });
+  });
```
ccharly, line 158: Same here, this could be a `it.each`

dawnseeker8: have used it.each to rewrite the tests.

ccharly (approved):

PR #302, opened 2025-06-26
feat: add multichain-account-api

ccharly:
Draft implementation of the new multichain accounts API and some adapters around it to help with the integration.

### DIFF
--- a/packages/multichain-account-api/CHANGELOG.md
+++ b/packages/multichain-account-api/CHANGELOG.md
@@ -1,0 +1,10 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+[Unreleased]: https://github.com/MetaMask/accounts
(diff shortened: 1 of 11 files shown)


ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

ccharly:
@metamaskbot publish-preview

packages/multichain-account-api/src/adapters.ts
```diff
@@ -0,0 +1,292 @@
...
+          });
+        }
+
+        // We've got all the accounts now, we can create our multichain account.
+        multichainAccounts.push(this.#createMultichainAccount(groupIndex));
+
+        // We have accounts, we need to check the next index.
+        groupIndex += 1;
+      }
+    } while (accounts.length);
```
ccharly, line 288: Question: Maybe we should just run discovery with the providers that HAS discovered something? (e.g Provider1 found something on index 0, but Provider2 did not, then we continue with index 1 for Provider1 but without Provider2?

If we do that, then we can use the "alignment mechanism" at the end of the discovery to re-align all accounts accross the multichain account.

packages/multichain-account-api/src/adapters.ts
```diff
@@ -0,0 +1,292 @@
...
+
+      if (accounts.length) {
+        // We only create missing accounts if one of the provider has discovered
+        // and created accounts.
+        for (const provider of missingProviders) {
+          await provider.createAccounts({
+            entropySource: this.#entropySource,
+            groupIndex,
+          });
+        }
```
ccharly, line 280: Move to async/parallel.

ccharly:
@metamaskbot publish-preview

PR #308, opened 2025-07-11
feat(keyring-api): test possible new account support

ulissesferreira:


packages/keyring-api/src/index.ts
```diff
@@ -2,6 +2,7 @@ export * from './api';
 export * from './btc';
 export * from './sol';
 export * from './eth';
+export * from './tron';
```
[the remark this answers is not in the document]
ulissesferreira: I wasn't sure if we prefered symbol or readability, I agree with `trx` let's maintain the pattern

packages/keyring-api/src/api/account.ts
```diff
@@ -30,6 +30,13 @@ export enum SolAccountType {
   DataAccount = 'solana:data-account',
 }
 
+/**
+ * Supported Tron account types.
+ */
+export enum TronAccountType {
+  DataAccount = 'tron:data-account',
```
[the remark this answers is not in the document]
ulissesferreira: Now that I have though about it more I agree, Tron is much more similar to the EVM world than Solana. It does not make sense to go for `data-account` I just did a poor search job on their documentation. To whoever sees this thread in the future: https://developers.tron.network/docs/account#account-types

packages/keyring-api/src/api/account.ts
```diff
@@ -30,6 +30,13 @@ export enum SolAccountType {
   DataAccount = 'solana:data-account',
 }
 
+/**
+ * Supported Tron account types.
```
ccharly, line 34:
```suggestion
 * Supported TRON account types.
```

packages/keyring-api/src/trx/types.ts
```diff
@@ -0,0 +1,54 @@
...
+  KeyringAccountStruct,
+  TronAccountType,
+} from '../api';
+
+/**
+ * TRON addresses are Base58-encoded strings that are exactly 34 characters long
+ * and start with the letter 'T'.
+ * Mainnet addresses begin with 'T'.
+ */
+export const TronAddressStruct = definePattern(
```
ccharly, line 17: Should probably be `Trx*` for everything that's TRON related in this file.

packages/keyring-api/src/trx/types.ts
```diff
@@ -0,0 +1,54 @@
...
+import {
+  CaipChainIdStruct,
+  KeyringAccountStruct,
+  TronAccountType,
+} from '../api';
+
+/**
+ * TRON addresses are Base58-encoded strings that are exactly 34 characters long
+ * and start with the letter 'T'.
+ * Mainnet addresses begin with 'T'.
```
danroc, line 15: Can you also document the pattern of testnet account addresses?

ulissesferreira: Removed that last line as it was misleading. I looked into the testnets and they all follow the same standard 👍

danroc (commented):
LGTM. My only comment was about the chain IDs, but we already addressed that with @ulissesferreira after some internal discussion.

ccharly (approved):
LGTM!

PR #313, opened 2025-07-15
build: use major `keyring-api` version for peer deps

ccharly:
We were wrongly using `workspace:^` when referencing the `@metamask/keyring-api` package as a peer dep. Thus, any small bump to that package would have make the packages (that depends on it) to be major bump every time (given that bumping a peer dep, results in a breaking change for the end-consumer).

We instead reference the current major version, thus allowing to not introduce breaking changes for every `@metamask/keyring-api` releases.

### DIFF
--- a/.syncpackrc
+++ b/.syncpackrc
@@ -29,7 +29,7 @@
   "versionGroups": [
     {
       "label": "use workspace version of packages in the monorepo",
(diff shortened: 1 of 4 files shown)


Gudahtt:
Great idea! You could also consider using Yarn Constraints to ensure we don't forget to update this. We have this setup already in core.

PR #314, opened 2025-07-16
fix(keyring-eth-ledger-bridge): increase ledger iframe timeout

Akaryatrh:
### What is the current state of things and why does it need to change?
We get a lot of Sentry reports about a ledger iframe timeout error

### What is the solution your changes offer and how does it work?
As a first mitigation, we will increase the timeout duration and then evaluate if we need to come up with a more robust solution. On a side track, we are also making [ledger transport initialization non-blocking](https://github.com/MetaMask/metamask-extension/pull/34574).

### Are there any issues or other links reviewers should consult to understand this pull request better?

Relates to: https://github.com/MetaMask/metamask-extension/issues/33176


Akaryatrh:
@metamaskbot publish-preview

PR #315, opened 2025-07-16
feat: add multichain account/wallet (readonly) support

ccharly:
Adding support for multichain accounts and wallets.

For now, we only support "read operations", meaning we can only construct multichain accounts and wallets based on existing keyring/internal accounts.


ccharly:
> ### Bug: Infinite Loop and Non-Contiguous Index Handling
> The `do-while` loop in the `MultichainAccountWalletAdapter` constructor lacks an upper bound, risking an infinite loop if `multichainAccount.hasAccounts()` consistently returns true (e.g., for HD wallets). Additionally, it incorrectly assumes account group indices are contiguous; the loop terminates upon encountering a `groupIndex` with no accounts, preventing the discovery of valid accounts at higher, non-contiguous indices (e.g., accounts at index 0 and 2, but not 1, will cause index 2 to be missed).

This is a valid concern. I did add a comment to explain that we were not covering those cases for the moment.

We could tweak the logic a bit on the account providers to avoid this in the future. For now, we can keep this simple.

packages/account-api/src/api/multichain/account.ts
```diff
@@ -0,0 +1,253 @@
...
+  return `${walletId}/${groupIndex}`;
+}
+
+/**
+ * Gets the multichain account index from an account group ID.
+ *
+ * @param groupId - Account group ID.
+ * @returns The multichain account index if extractable, undefined otherwise.
+ */
+export function getGroupIndexFromAccountGroupId(
```
danroc, line 241: Maybe this change since not all `AccountGroup` will have a group index.

```suggestion
export function getGroupIndexFromMultichainAccountId(
```

packages/account-api/src/api/provider.ts
```diff
@@ -0,0 +1,30 @@
...
+ * an account provider is reponsible of providing accounts to an account group.
+ */
+export type AccountGroupProvider<Account extends KeyringAccount> = {
+  /**
+   * Gets an account from its ID.
+   *
+   * NOTE: Assuming getting an account can never fail and will always be
+   * invoked with a valid account ID coming from a `getAccounts` response.
+   *
+   * @param id - The "blochain" account ID.
```
danroc, line 13:
```suggestion
   * @param id - The account ID.
```

packages/account-api/src/api/multichain/account.ts
```diff
@@ -0,0 +1,258 @@
...
+
+const MULTICHAIN_ACCOUNT_ID_REGEX = new RegExp(
+  `^${AccountWalletCategory.Entropy}:.*/(?<groupIndex>\\d+)$`,
+  'u',
+);
+
+/**
+ * Selector to query a specific account based on some criteria.
+ */
+export type MultichainAccountSelector<Account extends KeyringAccount> = {
```
danroc, line 20: Maybe is should be called `KeyringAccountSelector` since it will be used to "select" `KeyringAccount` objects that are part of an `AccountGroup`.

ccharly: 🤔 Maybe I could go with `AccountSelector` then... Given that we mostly use a generic type `Account` everywhere (which MUST BE a `KeyringAccount` yes)

WDYT?

ccharly:
@metamaskbot publish-preview

danroc (approved):

PR #320, opened 2025-07-21
feat(account-api): add reverse mapping + `AccountProvider` events

ccharly:
Implement reverse-mapping + event mechanism for the `AccountProvider`.

This also solve the issue that a wallet might not know when a new multichain account got added to the wallet.

For now, we omitted any use of `:accountRemoved` event since we don't plan to remove those accounts for the moment.

### DIFF
--- a/packages/account-api/package.json
+++ b/packages/account-api/package.json
@@ -46,6 +46,7 @@
     "test:watch": "jest --watch"
   },
   "dependencies": {
+    "@metamask/base-controller": "^8.0.1",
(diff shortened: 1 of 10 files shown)


packages/account-api/package.json
```diff
@@ -46,6 +46,7 @@
     "test:watch": "jest --watch"
   },
   "dependencies": {
+    "@metamask/base-controller": "^8.0.1",
```
ccharly, line 49: Required for the `Messenger`, but we will change that to `@metamask/messenger` once available.

PR #321, opened 2025-07-22
feat(account-api): enforce `Bip44Account` for `MultichainAccount*` + add `MultichainAccount*.sync`

ccharly:
To avoid adding an event-based mechanism to the `AccountProvider` we (for now) rely on external wallet sync's (which will be handle by the new `MultichainAccountService`.

This allow existing wallets to "find" new accounts and potentially create new missing multichan accounts.

Also, we now enforce `Bip44Account<Account>` for all `Multichain*` classes. This allows us to use `options.entropy.groupIndex` and quickly index multichain accounts based on that, thus, forcing the `AccountProvider` to only provide valid BIP-44 accounts too.

### DIFF
--- a/packages/account-api/src/api/multichain/account.ts
+++ b/packages/account-api/src/api/multichain/account.ts
@@ -35,9 +35,9 @@ export class MultichainAccount<Account extends Bip44Account<KeyringAccount>>
 
   readonly #providers: AccountProvider<Account>[];
 
-  readonly #accounts: Map<AccountProvider<Account>, Account['id'][]>;
+  readonly #providerToAccounts: Map<AccountProvider<Account>, Account['id'][]>;
 
-  readonly #reverse: Map<Account['id'], AccountProvider<Account>>;
+  readonly #accountToProvider: Map<Account['id'], AccountProvider<Account>>;
 
   constructor({
     groupIndex,
@@ -52,8 +52,8 @@ export class MultichainAccount<Account extends Bip44Account<KeyringAccount>>
     this.#index = groupIndex;
     this.#wallet = wallet;
     this.#providers = providers;
-    this.#accounts = new Map();
-    this.#reverse = new Map();
+    this.#providerToAccounts = new Map();
+    this.#accountToProvider = new Map();
 
     this.sync();
   }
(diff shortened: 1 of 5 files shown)


packages/account-api/src/api/multichain/account.ts
```diff
@@ -26,7 +24,7 @@ export type MultichainAccountId = `${MultichainAccountWalletId}/${number}`; // U
 /**
  * A multichain account that holds multiple accounts.
  */
-export class MultichainAccount<Account extends KeyringAccount>
+export class MultichainAccount<Account extends Bip44Account<KeyringAccount>>
```
ccharly, line 27: Enforcing `Bip44Account` here, allows us to use the new typed options pretty easily (the `options.entropy.groupIndex` is guaranteed to be defined with this)

Also, this forces the `AccountProvider` to only provide `Bip44Account` which should also prevent unwanted accounts during filtering.

packages/account-api/src/api/multichain/account.ts
```diff
@@ -37,6 +35,10 @@ export class MultichainAccount<Account extends KeyringAccount>
 
   readonly #providers: AccountProvider<Account>[];
 
+  readonly #accounts: Map<AccountProvider<Account>, Account['id'][]>;
+
+  readonly #reverse: Map<Account['id'], AccountProvider<Account>>;
+
```
hmalik88, line 41:
```suggestion
  readonly #providerToAccounts: Map<AccountProvider<Account>, Account['id'][]>;

  readonly #accountToProvider: Map<Account['id'], AccountProvider<Account>>;

```

packages/account-api/src/index.test.ts
```diff
@@ -41,18 +41,20 @@ import {
   isBip44Account,
 } from './api';
 
+type MockedAccount = Bip44Account<InternalAccount>;
+
 const mockEntropySource = 'mock-entropy-source';
 
-const mockAccountOptions: KeyringAccountOptions = {
+const mockAccountOptions = {
```
hmalik88, line 48: Why isn't this typed anymore?

hmalik88 (approved):

PR #333, opened 2025-07-30
refactor(account-api)!: remove multichain account implementations

ccharly:
Removing the current multichain implementation and move it to the [`MultichainAccountService`](https://github.com/MetaMask/core/tree/main/packages/multichain-account-service).

We still keep some type definitions, so we can re-use them without having to depend on the service itself.

Lastly, the naming convention used for "multichain accounts" was a bit off compared to `AccountGroup`, so the name has been updated from `MultichainAccount*` to `MultichainAccountGroup*`.


ccharly:
@metamaskbot publish-preview

hmalik88 (approved):

PR #341, opened 2025-08-05
feat(account-api/mocks): add common mock account definitions + `MockAccountBuilder`

ccharly:
Adding some test utils in the `account-api`, so we can use them whenever we need some mocked accounts and use the `MockAccountBuilder`.


ccharly:
@metamaskbot publish-preview

PR #347, opened 2025-08-06
release: 63.0.0

ccharly:
# Description

This is the release candidate for version 62.0.0. See the CHANGELOGs for more details.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/accounts-monorepo",
-  "version": "62.0.0",
+  "version": "63.0.0",
   "private": true,
   "description": "Monorepo for MetaMask accounts related packages",
   "repository": {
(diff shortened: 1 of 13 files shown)


fabiobozzo (approved):
QQ: why do the changelogs of `keyring-*` packages other than keyring-api say

> Bump `@metamask/keyring-api` from `^19.0.0` to `^20.0.0`

Where is that bumped exactly @ccharly ? Do those packages depend on keyring-api? 🤔

ccharly:
> QQ: why do the changelogs of `keyring-*` packages other than keyring-api say
> 
> > Bump `@metamask/keyring-api` from `^19.0.0` to `^20.0.0`
> 
> Where is that bumped exactly @ccharly ? Do those packages depend on keyring-api? 🤔

So, actually we're using `workspace:^` when referencing local packages from this monorepo, so every time we make a release of those packages, it gets "replaced" by the current version of each packages.

So this kind of happens "implicitly" and it's pretty hard to track I must say... We might want to change that someday!

fabiobozzo (dismissed):

PR #349, opened 2025-08-12
feat: adds Tron support

zone-live:
We are adding Tron support to keyring-api.

### DIFF
--- a/packages/keyring-snap-bridge/src/migrations/v1.ts
+++ b/packages/keyring-snap-bridge/src/migrations/v1.ts
@@ -8,11 +8,7 @@ import {
   TrxScope,
   TrxAccountType,
 } from '@metamask/keyring-api';
-import type {
-  CaipChainId,
-  KeyringAccount,
-  KeyringAccountType,
-} from '@metamask/keyring-api';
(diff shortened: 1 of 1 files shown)


packages/keyring-snap-bridge/src/account.ts
```diff
@@ -95,5 +101,5 @@ export function transformAccount(
     : accountToTransform;
 
   // We still assert that the converted account is valid according to their account's type.
-  return assertKeyringAccount(account);
+  return assertKeyringAccount(account as { type: KeyringAccountType });
```
aganglada, line 104: why do we need to do `as` here?

packages/keyring-snap-bridge/src/migrations/v1.ts
```diff
@@ -102,7 +108,7 @@ export function transformAccountV1(
 
   // For all other non-EVM and ERC4337 Snap accounts, the `scopes` is required, and
   // each `*AccountStruct` should assert that automatically.
-  return assertKeyringAccount(accountV1);
+  return assertKeyringAccount(accountV1 as { type: KeyringAccountType });
```
aganglada, line 111: same here

shane-t (approved):

PR #357, opened 2025-08-27
fix: fingerprint derivation for `CryptoHDKey`

mikesposito:
In the case of `CryptoHDKey`, the fingerprint was being derived from the parent key, instead of the origin source fingerprint. This brings the [extension patch](https://github.com/MetaMask/metamask-extension/blob/60744f4d38df0e54e8925a08a73b0eb422dffc2d/.yarn/patches/%40metamask-eth-qr-keyring-npm-1.0.0-8a180e9772.patch) upstream to match the expected behavior.


Akaryatrh (approved):

packages/keyring-eth-qr/src/device.ts
```diff
@@ -144,7 +144,7 @@ export type DeviceOptions = {
 function getFingerprintFromSource(source: CryptoAccount | CryptoHDKey): string {
   return source instanceof CryptoAccount
     ? source.getMasterFingerprint()?.toString('hex')
-    : source.getParentFingerprint()?.toString('hex');
+    : source.getOrigin()?.getSourceFingerprint()?.toString('hex');
 }
```
ccharly, line 148: Small remark here, we use a lot of `?`, meaning the response could really be `undefined`, but the return type is set to `string`, I guess we assume it to always be defined, right?

mikesposito: Unfortunately I already merged the PR before reading this comment.

You are right, fingerprints may be undefined for some devices based on these optional chained calls, and we could probably change the behavior to something else; though for now they follow the original package behavior

PR #505, opened 2026-04-08
feat: add keyring state migration framework

danroc:
## Summary

- Adds a versioned migration framework to `@metamask/keyring-sdk` for evolving keyring serialized state
- Migrations run during `deserialize()` and wrap state in a `{ version, data }` envelope
- `defineMigration()` helper provides compile-time type binding between `migrate` and superstruct `schema`
- Includes developer documentation with integration examples

## Test plan

- [ ] `yarn workspace @metamask/keyring-sdk test` passes (85 tests, 100% coverage)
- [ ] `yarn workspace @metamask/keyring-sdk build` succeeds
- [ ] No lint errors

---

> [!NOTE]
> **Medium Risk**
> New library for transforming serialized keyring/vault state; bugs in future migrations could corrupt sensitive persisted data, though this PR only adds the framework and does not wire it into existing keyrings.
> 
> **Overview**
> Adds a **versioned migration framework** to `@metamask/keyring-sdk` so keyrings can evolve persisted serialized state during `deserialize()` without breaking older vaults.
> 
> Callers build a forward-only chain with `createMigrations().add(...)`, where each step’s position is its version. `apply(state)` treats legacy unversioned JSON as version 0, runs only pending steps, and returns a `{ version, data, migrated }` envelope (`VersionedState` / `MigrationResult`). Optional superstruct `inputSchema` / `outputSchema` validate step I/O; `isVersionedState` detects the envelope. The module is **exported from the package entry** and documented in `docs/migrations.md` (integration pattern for `serialize()` / `deserialize()`).
> 
> Includes unit tests, `tsd` type tests for chained step typing, and an **Unreleased** changelog entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 688b35e36d431584343cb291b6180fc009f804eb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/keyring-sdk/docs/migrations.md
```diff
@@ -0,0 +1,118 @@
...
+}
+```
+
+Unversioned state (vaults created before migration support was added) has no envelope. The framework treats it as version 0 and applies all migrations.
+
+## Key Concepts
+
+- **`schema`**: Validates the **output** of a migration at runtime.
+- **`inputSchema`**: (Optional) Validates the **input** before the `migrate` function is called.
+- **`defineMigration<Output, Input>`**: Ensures the `migrate` function's return type matches the `schema` type at compile time.
```
ccharly: I would have expected to have input first, then output? But I guess it's because we want to have the type for the right version "first"?

Like:
```ts
const migrations = defineMigrations(
  // version 1 -> HdStateV1
  defineMigration<HdStateV1, HdStateV0>({
    version: 1,
    inputSchema: HdStateV0Schema,
    schema: HdStateV1Schema,
    migrate: (state) => ({
      accountCount: state.numberOfAccounts,
      mnemonic: state.mnemonic,
      hdPath: state.hdPath,
    }),
  }),
  ...
```

packages/keyring-sdk/docs/migrations.md
```diff
@@ -0,0 +1,118 @@
...
+
+### 3. Implement in your Keyring
+
+```typescript
+import { applyMigrations, getLatestVersion } from '@metamask/keyring-sdk';
+import type { Json } from '@metamask/utils';
+
+class MyKeyring {
+  async deserialize(state: Json): Promise<void> {
+    const { data } = await applyMigrations(state, migrations);
```
ccharly: WDYT about renaming this just `migrate`?

packages/keyring-sdk/docs/migrations.md
```diff
@@ -0,0 +1,118 @@
...
+const migrations = defineMigrations(
+  defineMigration<HdStateV1, HdStateV0>({
+    version: 1,
+    inputSchema: HdStateV0Schema,
+    schema: HdStateV1Schema,
+    migrate: (state) => ({
+      accountCount: state.numberOfAccounts,
+      mnemonic: state.mnemonic,
+      hdPath: state.hdPath,
+    }),
```
ccharly: Maybe we could auto-`Infer` input and output based on `inputSchema` and `schema`? 🤔

packages/keyring-sdk/docs/migrations.md
```diff
@@ -0,0 +1,118 @@
...
+});
+type HdStateV1 = Infer<typeof HdStateV1Schema>;
+
+const HdStateV2Schema = object({
+  accountCount: number(),
+  mnemonic: array(number()),
+  hdPath: string(),
+  createdAt: number(), // new field
+});
+type HdStateV2 = Infer<typeof HdStateV2Schema>;
```
ccharly, line 57: Nit: But would be nice to enforce that all schemas are JSON-compatible too (not sure we have something ready for this).

Maybe we don't need to enforce this in the `superstruct` itself, but in the `defineMigration*` we could make sure `T extends Json`? Just to avoid allowing state that could not be serialized to JSON?

danroc: It's already the case (at least in compile-time), `Input` and `Output` must extend `Json`.

packages/keyring-sdk/docs/migrations.md
danroc: It's already the case, they aren't necessary in this case, I can remove the types from the example.

packages/keyring-sdk/docs/migrations.md
danroc: `migrate` is already a field of `KeyringMigration`, I didn't re-use it because I thought it could be confusing.

packages/keyring-sdk/docs/migrations.md
danroc: It's intentional since `Output` is the primary type. `Input` can be omitted and it will default to `Json`.

packages/keyring-sdk/src/migration.test-d.ts
```diff
@@ -0,0 +1,53 @@
...
+      expectType<number>(data.count);
+      return data;
+    },
+  });
+
+// A step's `migrate` must accept the previous step's output shape.
+createMigrations()
+  .add({ migrate: (): { count: number } => ({ count: 1 }) })
+  // @ts-expect-error [test] `data` is `{ count: number }`, not `{ label: string }`.
+  .add({ migrate: (data: { label: string }) => data.label });
```
ccharly: Is that expected to use `@ts-expect-error` here?

We could use this to check the inferred input type maybe:

```suggestion
  .add({ migrate: (data) => {
    expectType<{ count: number }>(data);
    return data;
  }});
```

ccharly: But maybe, using `@ts-expect-error` is pretty self-describing already, and that would also fail in case this expression suddenly starts to succeed (I think the linter would see that as an error)

So maybe we can keep that!

ccharly (dismissed):
LGTM! Left 2 open question on the typing tests, but after 2nd thoughts, I think the current approach might be better.

I'll let you decide 😄

packages/keyring-sdk/docs/migrations.md
danroc: Actually, I reversed the order, now it's `Input, Output`

ccharly (approved):

PR #507, opened 2026-04-10
feat(keyring-snap-bridge): moving v1 `SnapKeyring` implementation to `SnapKeyringV1`

ccharly:
Following-up on:
- https://github.com/MetaMask/accounts/pull/501

The previous PR moved the v2 implementation into a proper class (per-Snaps), to slowly extract the monolith logic into more logical piece that we will re-use when splitting the existing `SnapKeyring`.

This new PR now also move the "v1 logic" out of it and move it in a dedicated keyring class (per-Snaps). The `SnapKeyringV2` now extends this v1, making the `SnapKeyringV2` compatible with the v2 interface AND also supports the old one (allowing us to do a progressive update with the new incoming `withKeyringV2`).

The remaining part on the `SnapKeyring` acts now as a coordinator/proxy around those new keyring objects. In the future, we would want to move this in a dedicated service (e.g the `SnapAccountService`) and each keyrings would live as usual, on the `KeyringController`, see:

Screenshot 2026-04-09 at 14 49 23

---

> [!NOTE]
> **Medium Risk**
> Refactors core Snap keyring message routing and request/account lifecycle handling into new per-snap classes, which could affect account events and pending request resolution across snaps. Adds teardown behavior (`destroy`) that rejects in-flight async requests when a snap keyring is removed or state is reloaded.
> 
> **Overview**
> **Refactors Snap keyring implementation into per-snap classes.** Introduces `SnapKeyringV1` (legacy event-driven flow + request lifecycle) and makes `SnapKeyringV2` extend it, so v2 batch APIs and v1 events share the same per-snap registry/client.
> 
> `SnapKeyring` is simplified into a coordinator/proxy: it lazily creates per-snap keyrings (only on `AccountCreated`), routes signing/requests to the owning snap keyring, and now *throws* when receiving messages for unknown snap keyrings. It also adds explicit per-snap teardown via `destroy()` (used on last-account removal and before `deserialize`) to reject pending async requests, and serializes `createAccounts` across snaps with a global mutex.
> 
> Tests are updated to reflect the new unknown-snap error behavior and to assert pending requests are rejected when the last account/keyring is removed; `SnapKeyringV1` is exported from `index.ts` and the changelog is updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 32951e726658e9ba97581900d395f85b4698cceb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/keyring-snap-bridge/src/SnapKeyringV2.ts
```diff
@@ -277,17 +195,9 @@ export class SnapKeyringV2 implements KeyringV2 {
           // Make sure to only delete accounts that were not part of the keyring state.
           if (!this.#getExistingAccount(snapAccount)) {
             try {
-              await this.#callbacks.deleteSnapAccount(snapAccount.id);
-            } catch (rollbackError) {
-              // Best-effort rollback; log snap-side failures for observability.
```
hmalik88, line 282: How come we're removing the log?

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -915,12 +290,11 @@ export class SnapKeyring {
     this.#snapKeyrings.clear();
     this.#accountIndex.clear();
 
-    // Rebuild per-snap wrappers. Each wrapper handles its own validation
-    // and migration internally.
+    // Rebuild per-snap entries. Migrations run inside v2.deserialize().
     for (const [snapId, accounts] of bySnap) {
-      const wrapper = this.#getOrCreateKeyringV2(snapId);
-      await wrapper.deserialize({ snapId, accounts });
-      this.#removeSnapKeyringIfEmpty(snapId);
```
hmalik88, line 928: why do we remove this?

packages/keyring-snap-bridge/src/SnapKeyringV2.ts
ccharly: Since I based my change on a previous version of your PR, I did overwrite too much stuff with `git checkout --ours` 🙈

I'll revert that, since your logic was already good!

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -915,12 +290,11 @@ export class SnapKeyring {
...
+      const entry = this.#getOrCreateKeyring(snapId);
```
hmalik88: Can we use `keyring` instead if you don't want to use wrapper, `entry` feels a little less intuitive

packages/keyring-snap-bridge/src/SnapKeyringV2.ts
ccharly: Ok reverted that back to normal, sorry!

packages/keyring-snap-bridge/src/SnapKeyring.ts
ccharly: Sure, I did a first pass initially, but looks like I missed A LOT of them, changing that now!

packages/keyring-snap-bridge/src/SnapKeyringV2.ts
ccharly: https://github.com/MetaMask/accounts/pull/507/commits/4a3be50e73ad4cacc689c8583f32abb8980aa6d6

packages/keyring-snap-bridge/src/SnapKeyring.ts
ccharly: https://github.com/MetaMask/accounts/pull/507/commits/29323412cb550520bd7d963c957cf19f04978ee3

packages/keyring-snap-bridge/src/SnapKeyringV2.ts
```diff
@@ -384,10 +312,10 @@ export class SnapKeyringV2 implements KeyringV2 {
    * @param account - The account to add or update.
    */
   setAccount(account: KeyringAccount): void {
-    const isNew = !this.#registry.has(account.id);
-    this.#registry.set(account);
+    const isNew = !this.registry.has(account.id);
+    this.registry.set(account);
     if (isNew) {
-      this.#onRegister(account.id);
+      this.#callbacks.onRegister?.(account.id);
```
hmalik88, line 356: why do we need the optional chaining here?

packages/keyring-snap-bridge/src/SnapKeyring.ts
ccharly: Another thing that got lost with my merge, truly sorry.. I should have tried to rebase a bit harder than that..

packages/keyring-snap-bridge/src/SnapKeyringV2.ts
```diff
@@ -400,11 +328,11 @@ export class SnapKeyringV2 implements KeyringV2 {
...
    */
   removeAccount(id: AccountId): boolean {
-    if (!this.#registry.has(id)) {
+    if (!this.registry.has(id)) {
       return false;
     }
-    this.#registry.delete(id);
-    this.#onUnregister(id);
+    this.registry.delete(id);
+    this.#callbacks.onUnregister?.(id);
```
hmalik88, line 373: Same as above

packages/keyring-snap-bridge/src/SnapKeyring.ts
ccharly: https://github.com/MetaMask/accounts/pull/507/commits/1ad9ee6c547e0219166d8cb1298c78230b8a7fc9

packages/keyring-snap-bridge/src/SnapKeyringV2.ts
ccharly: Ah yes, I made it optional so we can use this for the "legacy" `SnapKeyring`, but we won't have to use it when we'll use `SnapKeyringV2` as a "standalone" keyring (after the split).

Ultimately, this will be gone entirely.

I also thought of not having those and make multiple call to `lookup*` to fetch the accounts when we needed.

But I went with the least minimum changes for this part.

packages/keyring-snap-bridge/src/SnapKeyringV2.ts
ccharly: See https://github.com/MetaMask/accounts/pull/507#discussion_r3065339632

ccharly:
@metamaskbot publish-preview

packages/keyring-snap-bridge/src/SnapKeyringV2.ts
```diff
@@ -50,82 +49,40 @@ export const SnapKeyringV2StateStruct = object({
...
-   */
-  onUnregister: (accountId: AccountId) => void;
   /**
-   * Callbacks for snap communication and parent coordination.
+   * Mutex that serializes `createAccounts` calls on this snap instance.
+   * Owned here so that each `SnapKeyringV2` is fully self-contained.
    */
-  callbacks: SnapKeyringV2Callbacks;
-};
+  readonly #lock: Mutex;
```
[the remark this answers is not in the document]
ccharly: Since we're moving away of having a "monolithic" keyring for Snap accounts, a global lock might be more difficult with the new "split" version of keyrings.

But for now, I'll re-introduce it as an optional lock, so we can decide to delegate this to some other components if needed:

https://github.com/MetaMask/accounts/pull/507/commits/8edff3ccf75bf6a4e9e202adb54f339727e712df

packages/keyring-snap-bridge/src/SnapKeyringV1.ts
```diff
@@ -0,0 +1,1123 @@
...
+   * @returns The existing account if found, `undefined` otherwise.
+   */
+  #getExistingAccount(account: KeyringAccount): KeyringAccount | undefined {
+    const address = normalizeAccountAddress(account);
+    const existing = this.registry.get(account.id);
+    if (existing && normalizeAccountAddress(existing) === address) {
+      return existing;
+    }
+    return undefined;
+  }
```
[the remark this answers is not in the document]
ccharly: https://github.com/MetaMask/accounts/pull/507/commits/08beab5c108049a0ae437f2a5c9b4f7ff990d11c

packages/keyring-snap-bridge/src/SnapKeyringV1.ts
```diff
@@ -0,0 +1,1123 @@
...
+                // FIXME: There's a potential race condition here, if the Snap did
+                // not persist the account yet (this should mostly be for older Snaps).
+                this.registry.delete(account.id);
+                this.#callbacks.onUnregister?.(account.id);
+                await this.client
+                  .deleteAccount(account.id)
+                  /* istanbul ignore next */
+                  .catch(() => {
+                    // Swallow — best-effort cleanup.
+                  });
```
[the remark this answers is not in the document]
ccharly: https://github.com/MetaMask/accounts/pull/507/commits/71494179497b5f514858f09286e5c46bed70222e

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -240,124 +146,86 @@ export class SnapKeyring {
...
+   * @returns The SnapKeyringEntry for the given Snap.
    */
-  #getOrCreateKeyringV2(snapId: SnapId): SnapKeyringV2 {
+  #getOrCreateKeyring(snapId: SnapId): SnapKeyringV2 {
     let keyring = this.#snapKeyrings.get(snapId);
     if (!keyring) {
-      const client = new KeyringInternalSnapClient({
-        messenger: this.#messenger,
-        snapId,
-      });
```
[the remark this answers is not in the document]
ccharly: https://github.com/MetaMask/accounts/pull/507/commits/4ef4b6c13c60518667ff8d34d7a76bdd674abf35

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -392,435 +260,15 @@ export class SnapKeyring {
...
-   * Handle an Request Approved event from a Snap.
-   *
-   * @param snapId - Snap ID.
-   * @param message - Event message.
-   * @returns `null`.
-   */
-  async #handleRequestApproved(
-    snapId: SnapId,
-    message: SnapMessage,
-  ): Promise<null> {
```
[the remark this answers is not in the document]
ccharly: That's a problem we already had initially with the current `SnapKeyring`. Those requests are not "persisted" meaning that every time the keyring got re-created, we we were losing them (after a lock/unlock, keyring instances are getting re-created for example).

So it's not "worst" than what we had initially nor better.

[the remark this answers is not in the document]
ccharly: Though, having a `destroy` that clears pending requests is a good idea still. I added that:

https://github.com/MetaMask/accounts/pull/507/commits/32951e726658e9ba97581900d395f85b4698cceb

packages/keyring-snap-bridge/src/SnapKeyringV1.ts
```diff
@@ -0,0 +1,1146 @@
...
+                      deleteError,
+                    );
+                  });
+
+                // This allows the MetaMask client to be "notified" that something went
+                // wrong with the Snap keyring. (e.g. useful to display account creation
+                // error dialogs).
+                onceSaved.reject(error);
+              },
+            );
```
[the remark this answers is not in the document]
ccharly: `saveState` should never really fail. And handling the keyring removal inside the keyring itself (in this async flow) would be a real challenge.

IMO, that's acceptable, since and empty keyring has no real bad side-effects!

hmalik88 (approved):

PR #517, opened 2026-04-15
chore: update tooling (same as `core`)

ccharly:
Using new tooling (same as `core`).

---

> [!NOTE]
> **Medium Risk**
> Tooling changes upgrade ESLint to v9 and replace Prettier-driven misc linting with `oxfmt`, which can change lint/format outputs and CI behavior across the monorepo. Functional code changes are minimal (mostly import/type-only cleanups), but review should confirm the new lint/suppression setup matches current expectations.
> 
> **Overview**
> Switches from legacy `.eslintrc.js` to an ESLint v9 flat config in `eslint.config.mjs`, including `projectService` TypeScript parsing, targeted overrides, and a new `eslint-suppressions.json` for managed rule suppressions.
> 
> Adopts `oxfmt` for misc formatting (`.oxfmtrc.json`) and updates root scripts (e.g., `lint:*`, `create-release-branch`, build cleanup), while upgrading core tooling deps (ESLint 9, Prettier 3, `@metamask/eslint-config` v15) and dropping `syncpack`-based dependency mismatch checks.
> 
> Across packages, normalizes `package.json` publish metadata (`files`, `main`/`types`, `publishConfig`, `tsd` config) and applies mostly non-functional code churn (import ordering, `import type`, removing inline eslint-disable comments, small readability tweaks).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 32dfc963b0b86baa65f8f7de1527205056816596. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ccharly:
@SocketSecurity ignore npm/@metamask/create-release-branch@4.1.4

This our package.

.eslintrc.js
ccharly, line 1: This got replaced by `eslint.config.mjs` now (for eslint 9)

.oxfmtrc.json
ccharly, line 1: Same as `core`

.prettierrc.js
ccharly, line 1: Same as `core`

eslint-suppressions.json
ccharly, line 1: Also imported from `core` + new feature from eslint 9 (we'll need to fix those existing lint errors, but for now I wanted to keep this """as small as possible""")

eslint.config.mjs
ccharly, line 1: Converted from our initial `.eslintrc.js`

package.json
```diff
@@ -7,21 +7,25 @@
...
     "packages/*"
   ],
+  "files": [],
   "scripts": {
     "build": "ts-bridge --project tsconfig.build.json",
     "build:clean": "yarn build --clean",
     "build:docs": "yarn foreach build:docs",
+    "build:only-clean": "rimraf -g 'packages/*/dist'",
+    "build:types": "tsc --build tsconfig.build.json --verbose",
+    "create-release-branch": "create-release-branch --formatter oxfmt",
```
ccharly, line 20: That's new, but that's what `core` does use and since we're moving away from `prettier`, we need to override this command to automatically use `oxfmt`

package.json
```diff
@@ -7,21 +7,25 @@
...
     "lint": "yarn lint:eslint && yarn lint:misc --check && yarn constraints && yarn lint:dependencies && yarn lint:readme",
     "lint:dependencies": "yarn foreach depcheck && yarn dedupe --check && syncpack list-mismatches",
     "lint:dependencies:fix": "yarn foreach depcheck && yarn dedupe && syncpack fix-mismatches",
-    "lint:eslint": "eslint . --cache --ext js,cjs,mjs,ts",
-    "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write && yarn constraints --fix && yarn lint:dependencies:fix",
-    "lint:misc": "prettier '**/*.json' '**/*.md' '!**/CHANGELOG.old.md' '**/*.yml' '!.yarnrc.yml' '!merged-packages/**' --ignore-path .gitignore",
+    "lint:eslint": "yarn build:only-clean && yarn eslint",
+    "lint:fix": "yarn lint:eslint --fix --prune-suppressions && yarn lint:misc --write && yarn constraints --fix && yarn lint:dependencies:fix",
+    "lint:misc": "oxfmt --ignore-path .gitignore",
+    "lint:misc:check": "yarn lint:misc --check",
```
ccharly, line 28: Re-used what we have in `core` too, removed `prettier` since we're moving away from it

NOTE: We still need `prettier` for a bunch of other things, but at least, `oxfmt` is the default formatting tool for the code now

package.json
```diff
@@ -33,61 +37,66 @@
...
-  "packageManager": "yarn@4.10.3",
   "engines": {
     "node": "^18.18 || >=20"
   },
+  "packageManager": "yarn@4.10.3",
   "lavamoat": {
     "allowScripts": {
-      "@lavamoat/preinstall-always-fail": false
+      "@lavamoat/preinstall-always-fail": false,
+      "eslint-plugin-import-x>unrs-resolver": false
```
ccharly, line 101: Was required for some an eslint plugin (we also have this on `core`)

packages/hw-wallet-sdk/src/hardware-error-mappings.ts
```diff
@@ -250,7 +250,6 @@ export const QR_WALLET_ERROR_MAPPINGS: Record<string, ErrorMapping> = {
   },
 };
 
-/* eslint-disable @typescript-eslint/naming-convention */
```
ccharly, line 253: Seems to be enabled now with the default eslint MetaMask config

ccharly:
Added `no-changelog` since it's purely internal.

ccharly:
@SocketSecurity ignore npm/express@4.22.1

It obviously has network access and we need it

ccharly:
@SocketSecurity ignore npm/@octokit/request@8.4.1

Same here, it's legit.

ccharly:
@SocketSecurity ignore npm/async-function@1.0.0
@SocketSecurity ignore npm/encodeurl@2.0.0
@SocketSecurity ignore npm/finalhandler@1.3.2
@SocketSecurity ignore npm/http-errors@2.0.1
@SocketSecurity ignore npm/path-to-regexp@0.1.13
@SocketSecurity ignore npm/statuses@2.0.2

Those 3 authors are co-maintainers of common packages together and are maintaining a lot of other JS/TS packages (they also have valid GH users). Seems good to me.

ccharly:
@SocketSecurity ignore npm/oxfmt@0.44.0

We already use that tool on `core`.

ccharly:
@SocketSecurity ignore npm/body-parser@1.20.4

Uses `http-error` to create HTTP errors, that's fine.

ccharly:
@SocketSecurity ignore npm/methods@1.1.2

Uses `http.METHOD` obvisouly, that's fine.

ccharly:
@SocketSecurity ignore npm/raw-body@2.5.3

Uses `http-error` to create HTTP errors, that's fine.

ccharly:
@SocketSecurity ignore npm/send@0.19.2

Uses `http-error` to create HTTP errors, that's fine.

ccharly:
@SocketSecurity ignore npm/unrs-resolver@1.11.1

The `child_process` is only used for `exec` to invoke `ldd` (to get info on the MUSL C library I guess), there's no way to inject any command on this `exec`, so we're good here.

ccharly:
@SocketSecurity ignore npm/debug@2.6.9

We already uses this dep for our debug loggers. The `net` module is indeed used to create a stdio stream over a socket, but we are not using this.

ccharly:
@SocketSecurity ignore npm/@tybys/wasm-util@0.10.1

This package uses `fetch` to load `.wasm` code like:
```ts
// using URL
const { module, instance } = await load(new URL('./file.wasm', import.meta.url), imports)
const { module, instance } = loadSync(new URL('./file.wasm', import.meta.url), imports)
```

This seems to be a pretty popular package AND that's a transitive dep from a dev dep. We're not using the package directly.

EDIT: Actually, I added resolutions to avoid pulling this dep now. Not sure if `ignore` is still registered on socket, but this would be nice to "un-ignore" this now, since we don't need to acknowledge this dep anymore for this PR

.syncpackrc
ccharly, line 1: Also decided to drop this, since we can achieve what we want with our new `yarn.config.cjs`

package.json
```diff
@@ -33,61 +37,68 @@
...
-    "eslint-plugin-promise": "^6.6.0",
+    "eslint": "^9.39.1",
+    "eslint-config-prettier": "^9.1.0",
+    "eslint-import-resolver-typescript": "^3.6.3",
+    "eslint-plugin-import-x": "^4.3.0",
+    "eslint-plugin-jest": "^28.8.3",
+    "eslint-plugin-jsdoc": "^50.2.4",
+    "eslint-plugin-n": "^17.10.3",
+    "eslint-plugin-prettier": "^5.2.1",
+    "eslint-plugin-promise": "^7.1.0",
```
ccharly, line 67: Used same range than `core`

package.json
```diff
@@ -33,61 +37,68 @@
...
+    "typescript": "~5.6.3",
+    "typescript-eslint": "^8.48.0"
+  },
+  "resolutions": {
+    "@types/node": "^20.12.12",
+    "@types/web": "^0.0.69",
+    "@typescript/lib-dom": "npm:@types/web@^0.0.69",
+    "axios@1.7.3": "^1.7.7",
+    "eslint-import-resolver-typescript": "<=3.7.0",
+    "eslint-plugin-import-x": "<=4.6.1",
```
ccharly, line 90: Added those resolutions to avoid pulling too much deps. This way we keep using similar versions than core and we avoid having Socket warn/block reports on some transitive deps!

danroc:
We can enable the rules that we manually disabled in the `eslint.config.mjs` and update the `eslint-suppressions.json` to ignore the _existing_ violations.

packages/keyring-eth-ledger-bridge/src/ledger-keyring.test.ts
```diff
@@ -1515,7 +1506,7 @@ describe('LedgerKeyring', function () {
         const result = await keyring.getAppConfiguration();
 
         expect(result).toStrictEqual(mockResponse);
-        // eslint-disable-next-line @typescript-eslint/unbound-method
+
```
danroc:
```suggestion
```

packages/keyring-eth-trezor/src/trezor-keyring.test.ts
```diff
@@ -452,7 +452,7 @@ describe('TrezorKeyring', function () {
         newFakeTx,
       );
       // ensure we get a new version transaction back
-      // eslint-disable-next-line @typescript-eslint/unbound-method
+
```
danroc:
```suggestion
```

packages/keyring-eth-trezor/src/trezor-keyring.test.ts
```diff
@@ -487,7 +487,7 @@ describe('TrezorKeyring', function () {
         contractDeploymentFakeTx,
       );
       // ensure we get a new version transaction back
-      // eslint-disable-next-line @typescript-eslint/unbound-method
+
```
danroc:
```suggestion
```

packages/keyring-eth-trezor/src/trezor-keyring.test.ts
```diff
@@ -692,12 +692,12 @@ describe('TrezorKeyring', function () {
           domain: {},
           message: {},
         },
-        // eslint-disable-next-line @typescript-eslint/naming-convention
+
```
danroc:
```suggestion
```

packages/keyring-eth-trezor/src/trezor-keyring.test.ts
```diff
@@ -692,12 +692,12 @@ describe('TrezorKeyring', function () {
...
         metamask_v4_compat: true,
-        // eslint-disable-next-line @typescript-eslint/naming-convention
+
```
danroc:
```suggestion
```

packages/keyring-eth-trezor/src/trezor-keyring.test.ts
```diff
@@ -692,12 +692,12 @@ describe('TrezorKeyring', function () {
...
         domain_separator_hash:
           '6192106f129ce05c9075d319c1fa6ea9b3ae37cbd0c1ef92e2be7137bb07baa1',
-        // eslint-disable-next-line @typescript-eslint/naming-convention
+
```
danroc:
```suggestion
```

packages/keyring-eth-trezor/src/trezor-keyring.test.ts
```diff
@@ -764,7 +764,7 @@ describe('TrezorKeyring', function () {
       perPage: 10,
     };
     const accountToUnlock = 1;
-    // eslint-disable-next-line @typescript-eslint/naming-convention
+
```
danroc:
```suggestion
```

packages/keyring-snap-bridge/src/SnapIdMap.ts
```diff
@@ -249,7 +248,7 @@ export class SnapIdMap<Value extends { snapId: SnapId }> {
    * @param value - Value of the element to add or update.
    * @returns The map itself.
    */
-  /* eslint-enable jsdoc/check-indentation */
+
```
danroc:
```suggestion
```

ccharly:
> We can enable the rules that we manually disabled in the `eslint.config.mjs` and update the `eslint-suppressions.json` to ignore the _existing_ violations.

Oh that's a very good point indeed 👍 let me try to do this now

.syncpackrc
danroc: After checking the `yarn.config.cjs` file, I’m a bit on the fence here: it’s nice to have the rules in a single place, but the Syncpack config is much more readable and easier to maintain, given its declarative nature compared to the programmatic approach of Yarn constraints.

eslint-suppressions.json
```diff
@@ -0,0 +1,263 @@
+{
```
david0xd, line 1: My advice is to fix these before merging, and avoid doing `eslint-supressions`. While there are little bit more of these ESLint issues, there are no too many of them that we cannot fix it here I think 🤔 

This doesn't seem to be a good pattern IMO.

eslint-suppressions.json
```diff
@@ -0,0 +1,263 @@
...
+    "@typescript-eslint/no-explicit-any": {
+      "count": 1
+    }
+  },
+  "packages/keyring-utils/src/typing.ts": {
+    "@typescript-eslint/no-explicit-any": {
+      "count": 2
+    }
+  }
+}
```
danroc: Nit: missing new line at the end of the file.

```suggestion
}

```

packages/hw-wallet-sdk/src/hardware-error-mappings.ts
```diff
@@ -464,4 +463,3 @@ export const TREZOR_ERROR_MAPPINGS: Record<string, ErrorMapping> = {
       'Your Trezor device failed to initialize. Please unlock it and try again.',
   },
 };
```
danroc, line 465: Another nit: add a new line at the end of the file.

```suggestion
};

```

eslint-suppressions.json
ccharly: That's the pattern we have in `core` actually 😅

I just didn't want to have this PR to blow up too much.

WDYT about moving forward with this (to secure the other tools being updated like `auto-changelog` for easier release) and then we can follow-up with a big `lint:fix` and remove all those exceptions?

While auto-fixing them should not be too hard. I'm not sure if this would have some undesired side-effects (I had to add explicit exceptions cause the auto-fixer was doing some very weird stuff).

.syncpackrc
ccharly: Well, that's not completely false but I'm just trying to keep using the same tooling as `core` and I feel like `syncpack` is good, but we can definitely live without it IMO

eslint-suppressions.json
david0xd: I'm not blocking this PR. You can go ahead with it this way if you wish. 

I just wanted to rise awareness of this stuff since I think it is anti-pattern and it is defeating the purpose of the ESLint and its configurations we have in place for number of reasons.

I think that "core" is wrong in its approach.

The purpose of the ESLint is to stop us from merging bad things, and by suppressing it and skipping its validation, we're ultimately ending up in contradiction.

IMO this is a sort of "tool migration" PR and it's not heavy with logical complexity. So it is fine to have many changes within it, from my perspective.

eslint-suppressions.json
ccharly: Should be outdated now (or at least, the current lint step worked 😅)

danroc (approved):

PR #519, opened 2026-04-17
fix(keyring-snap-bridge)!: set Snap ID during `deserialize`

ccharly:
Using the `snapId` in the constructor was working fine in the current setup, unfortunately this cannot be used with our keyring lifecycle.

The way the `KeyringController` creates keyring is:
- Creates the keyring with its builder `new Keyring(...)`
  * We cannot really pass dynamic options here
  * The Snap keyring v2 will be instantiated BY the `KeyringController` that have no knowledge about Snap IDs
- Call `deserialize` with some custom data
  * That's what we need to use to inject the Snap ID we need

Some thing like:
```ts
await controller.addNewKeyring(KeyringType.Snap /* v2 */, { snapId, accounts: { ... } });
```

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes keyring initialization semantics (snapId no longer provided at construction) and adds new async locking around per-snap keyring creation, which could affect runtime ordering and lifecycle edge-cases.
> 
> **Overview**
> **BREAKING:** `SnapKeyring`/`SnapKeyringV2` no longer accept `snapId` in the constructor; the snap identity is now *bound on first* `deserialize`, and re-deserializing with a different `snapId` throws.
> 
> The bridge’s per-snap keyring creation path (`#getOrCreateKeyring`) is now async and protected by a dedicated mutex, eagerly calling `deserialize({ snapId, accounts: {} })` before registering the new keyring to avoid races/duplicate keyrings under concurrent calls. Tests were updated to reflect the new initialization flow and to cover concurrent creation against a previously-unknown snap.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7dff76aaba17b4cae4653a89f2de270640e399ee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/keyring-snap-bridge/src/SnapKeyring.ts
+++ b/packages/keyring-snap-bridge/src/SnapKeyring.ts
@@ -127,6 +127,16 @@ export class SnapKeyring {
   readonly #lock: Mutex;
 
   /**
+   * Serializes lazy keyring initialization in {@link SnapKeyring.#getOrCreateKeyring}.
+   *
+   * Kept separate from {@link SnapKeyring.#lock} (which guards `createAccounts`
+   * uniqueness checks) to avoid mixing concerns. The lock is held only for the
+   * duration of a single keyring's initialization - released before the caller
+   * proceeds with its operation.
+   */
+  readonly #getOrCreateKeyringLock: Mutex;
+
+  /**
    * Create a new Snap keyring.
    *
    * @param options - Constructor options.
@@ -151,6 +161,7 @@ export class SnapKeyring {
     this.#callbacks = callbacks;
     this.#isAnyAccountTypeAllowed = isAnyAccountTypeAllowed;
     this.#lock = new Mutex();
+    this.#getOrCreateKeyringLock = new Mutex();
   }
 
   /**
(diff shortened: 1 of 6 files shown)


packages/keyring-snap-bridge/src/SnapKeyringV1.ts
```diff
@@ -202,23 +201,87 @@ export class SnapKeyringV1 {
...
     this.#options = new Map();
     this.#selectedAccounts = [];
   }
 
+  /**
+   * Assert that this keyring has been initialized (i.e. `snapId` is set).
+   *
+   * @throws If the keyring has not been initialized yet.
+   */
+  protected assertIsInitialized(): void {
```
hmalik88: Nit, but you can do something like this to avoid the casts above:

```ts
  #requireInitialized(): {
    snapId: SnapId;
    client: KeyringInternalSnapClient;
  } {
    if (this.#snapId === undefined || this.#client === undefined) {
      throw new Error('SnapKeyring has not been initialized: call deserialize() first');
    }
    return { snapId: this.#snapId, client: this.#client };
  }
```

```ts
  get snapId(): SnapId {
    return this.#requireInitialized().snapId;
  }

  protected get client(): KeyringInternalSnapClient {
    return this.#requireInitialized().client;
  }
```

ccharly: Looks neat, let me change that!

ccharly: Went for this: 58a8555d4b8a2a35da2bd315ca19b213715a2b3c

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -206,6 +210,8 @@ export class SnapKeyring {
       });
 
       this.#snapKeyrings.set(snapId, keyring);
+      // Initialize the keyring with its snap ID (no accounts yet).
+      await keyring.deserialize({ snapId, accounts: {} });
```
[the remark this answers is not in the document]
ccharly: https://github.com/MetaMask/accounts/pull/519/commits/d1311650ed232f61acfc2e3581591a5fcd5b3ba0

packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -205,6 +209,10 @@ export class SnapKeyring {
         },
       });
 
+      // Initialize the keyring with its snap ID (no accounts yet).
+      await keyring.deserialize({ snapId, accounts: {} });
+
+      // Now the keyring is fully-ready, we can add it to the map.
```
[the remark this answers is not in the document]
ccharly: The usual TOCTOU problem 🥹 I guess I'll add a mutex for this then!

hmalik88 (approved):

PR #531, opened 2026-04-27
feat: add `isUnlocked` callback to `SnapKeyringCallbacks`

hmalik88:
Added `isUnlocked` callback that checks if the `KeyringController` is unlocked as an extra safety mechanism for in flight account creation. 

Note: This does not eliminate entirely all lock errors as the controller might become locked after the `isUnlocked` check but this should eliminate most of the errors we've seen as they originated when we would make our `addressExists` which would fail because of the `KeyringController` being locked

---

> [!NOTE]
> **Medium Risk**
> Changes account-creation behavior in both v1 event handling and v2 batch creation to early-return when the controller is locked, and introduces a new required callback that consumers must implement.
> 
> **Overview**
> Adds a new required `isUnlocked` callback to `SnapKeyringCallbacks`/`SnapKeyringV1Callbacks` and wires it through the per-snap keyring wrapper.
> 
> When `isUnlocked` is false, the bridge now **defers inflight account creation**: v1 `notify:accountCreated` handling returns `null` without running validation/callbacks, and v2 `createAccounts` returns an empty array without calling the Snap client. Tests and the changelog are updated to cover and document this locked-controller deferral behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2c0f6ca585b6dc665bded1c984a552df42d429a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/keyring-snap-bridge/CHANGELOG.md
+++ b/packages/keyring-snap-bridge/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added `isUnlocked` callback as a safety mechanism for inflight account creation ([]())
+- Added `isUnlocked` callback as a safety mechanism for inflight account creation ([#531](https://github.com/MetaMask/accounts/pull/531))
 
 ## [22.0.0]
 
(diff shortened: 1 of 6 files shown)


hmalik88:
Closing as this fix wouldn't work for 3rd party snaps, leaving things as they are since in v2 the lock issue won't be a problem.

PR #532, opened 2026-04-28
feat(keyring-eth-lattice): add initial implementation

ccharly:
A new package to wrap the existing v1 `eth-lattice-keyring`. We'll then proceed with a proper v2 wrapper so we will be able to use this keyring with `withKeyringV2`.


ccharly:
@SocketSecurity ignore npm/eth-lattice-keyring@1.1.0

We already have acknowledged this package and we are using it in production, see: https://github.com/MetaMask/metamask-extension/pull/39225#issuecomment-3952034019

PR #539, opened 2026-05-04
feat(keyring-snap-bridge): add `SnapKeyringV2` v1 adapter

ccharly:
Adding a v1 adapter for v2 `SnapKeyring` (1 keyring per Snap). So we can use it with the current architecture of the `KeyringController` that constructs a v2 keyring instance from a v1 instance.

The idea is to use this "adapter" for v1 instances and call `.asV2()` on this same instance when the `KeyringController` initializes its v2 counterpart. This way, we share the same actual instance/logic/account-registry between the 2.

NOTE: No one will use the v1 instance directly, but I add it mostly to be compliant with the current `KeyringController` architecture AND use it in v2 contexts (using `:withKeyringV2`).

One big important note though, the vault is currently computed using v1 instances on the `KeyringController`, thus, we still need this extra layer for it to be part of the vault and migrate from the legacy 1 big Snap keyring -> N Snap keyrings (1 per Snap), see:
- https://github.com/MetaMask/core/blob/0d7aac0baca77af2d2c78d166dbe5b8539024561/packages/keyring-controller/src/KeyringController.ts#L2561-L2569
- https://github.com/MetaMask/core/blob/0d7aac0baca77af2d2c78d166dbe5b8539024561/packages/keyring-controller/src/KeyringController.ts#L705-L718
- https://github.com/MetaMask/core/blob/0d7aac0baca77af2d2c78d166dbe5b8539024561/packages/keyring-controller/src/KeyringController.ts#L3006-L3016

---

> [!NOTE]
> **Medium Risk**
> Adds a new compatibility wrapper that changes how per-snap `SnapKeyring` instances can be registered/consumed by v1 `KeyringController` flows, which could affect account address normalization and vault/serialization behavior if integrated incorrectly.
> 
> **Overview**
> **Adds `SnapKeyringV1Adapter` under the `/v2` export** to let v1 `EthKeyring` consumers interact with per-snap v2 `SnapKeyring` instances, while still exposing the underlying v2 object via `asV2()` so v1/v2 entries can share a single instance.
> 
> The adapter **converts v2 `getAccounts()` results into v1-style `string[]` addresses** (using `normalizeAccountAddress`), and otherwise delegates snap message handling, signing/user-op methods, lifecycle (`destroy`), and `serialize`/`deserialize` directly to the inner v2 keyring. `SnapKeyringOptions` is now exported and the changelog documents the new adapter.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7e0a3c6f0c5fccaf727bc885598fdd0eeb89dcd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/keyring-snap-bridge/CHANGELOG.md
+++ b/packages/keyring-snap-bridge/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `SnapKeyringV1Adapter` (v2 subpath) class, a V1-compatible adapter for a per-snap `SnapKeyring` (v2) instance ([#539](https://github.com/MetaMask/accounts/pull/539))
+  - This adapter allows `KeyringController` to interact with per-snap `SnapKeyring` (v2) instances using the legacy v1 `EthKeyring` interface (e.g. `getAccounts()` returning `Promise<string[]>`).
+  - The inner `SnapKeyring` (v2) instance is accessible via `asV2()`, allowing `KeyringController`'s `KeyringV2Builder` to share the same object as the v2 keyring entry with no state duplication.
+
 ## [22.0.1]
 
 ### Changed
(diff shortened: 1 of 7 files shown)


ccharly:
Closing in favor of a more generic v1 adapter:
- https://github.com/MetaMask/accounts/pull/543

PR #564, opened 2026-06-01
feat: add 7702 methods to ledger

montelaidev:
This PR adds the 7702 methods to the ledger keyring. The method itself is not usable until the dmk support has been added. 

## Examples

---

> [!NOTE]
> **Medium Risk**
> Changes hardware signing and signature recovery for personal_sign/typed data; EIP-7702 is new delegation signing surface, though iframe/mobile paths are not live yet.
> 
> **Overview**
> Adds **EIP-7702 delegation authorization signing** to the Ledger Ethereum keyring and wires it through the V2 account API.
> 
> `LedgerKeyring.signEip7702Authorization` unlocks the account, calls the bridge’s new `deviceSignDelegationAuthorization` (chain id, contract address without `0x`, nonce), builds a signature with **yParity** (`00`/`01`) instead of legacy `27`/`28`, and verifies with `recoverEIP7702Authorization`. The `LedgerBridge` interface gains delegation types and `deviceSignDelegationAuthorization`; **iframe and mobile bridges throw** until DMK/device support exists.
> 
> **personal_sign** and **typed data** now parse Ledger’s `v` via `#parseLedgerRecoveryParam` (decimal, bare hex like `'1b'`, `0x`/`0X` hex, numbers) instead of `parseInt(..., 10)`, fixing wrong recovery when the device returns hex `v`.
> 
> V2 `LedgerKeyring` advertises `EthKeyringMethod.SignEip7702Authorization` on EOA accounts. Changelog and Jest coverage thresholds are updated; extensive unit tests cover 7702 signing and `v` parsing edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c91b13d32b8241be5f695a2d0c5048ee4b005e03. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/keyring-eth-ledger-bridge/CHANGELOG.md
+++ b/packages/keyring-eth-ledger-bridge/CHANGELOG.md
@@ -10,8 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add EIP-7702 authorization signing to the Ledger keyring ([#564](https://github.com/MetaMask/accounts/pull/564))
-    - Not yet functional: both iframe and mobile bridges throw until DMK support lands.
-    - Fixes hex `v` parsing (`'1b'`) for personal_sign and typed data. 
+  - Not yet functional: both iframe and mobile bridges throw until DMK support lands.
+  - Fixes hex `v` parsing (`'1b'`) for personal_sign and typed data.
 
 ### Changed
 
--- a/packages/keyring-eth-ledger-bridge/CHANGELOG.md
+++ b/packages/keyring-eth-ledger-bridge/CHANGELOG.md
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added 7702 methods ([#564](https://github.com/MetaMask/accounts/pull/564))
+- Add EIP-7702 authorization signing to the Ledger keyring ([#564](https://github.com/MetaMask/accounts/pull/564))
+    - Not yet functional: both iframe and mobile bridges throw until DMK support lands.
+    - Fixes hex `v` parsing (`'1b'`) for personal_sign and typed data. 
 
 ### Changed
 
--- a/packages/keyring-eth-ledger-bridge/CHANGELOG.md
+++ b/packages/keyring-eth-ledger-bridge/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added 7702 methods ([#564](https://github.com/MetaMask/accounts/pull/564))
+
 ### Changed
 
 - Bump `@metamask/keyring-api` from `^23.1.0` to `^23.2.0` ([#562](https://github.com/MetaMask/accounts/pull/562))
--- a/packages/keyring-eth-ledger-bridge/src/ledger-bridge.ts
+++ b/packages/keyring-eth-ledger-bridge/src/ledger-bridge.ts
@@ -32,11 +32,9 @@ export type LedgerSignDelegationAuthorizationParams = {
   nonce: number;
 };
 
-export type LedgerSignDelegationAuthorizationResponse = {
-  v: string;
-  r: string;
-  s: string;
-};
+export type LedgerSignDelegationAuthorizationResponse = Awaited<
+  ReturnType<LedgerHwAppEth['signTransaction']>
+>;
 
 export type GetAppNameAndVersionResponse = {
   appName: string;
--- a/packages/keyring-eth-ledger-bridge/src/ledger-bridge.ts
+++ b/packages/keyring-eth-ledger-bridge/src/ledger-bridge.ts
@@ -25,6 +25,19 @@ export type LedgerSignTypedDataResponse = Awaited<
   ReturnType<LedgerHwAppEth['signEIP712HashedMessage']>
 >;
 
+export type LedgerSignDelegationAuthorizationParams = {
+  hdPath: string;
+  chainId: number;
+  contractAddress: string;
+  nonce: number;
+};
+
+export type LedgerSignDelegationAuthorizationResponse = {
+  v: string;
+  r: string;
+  s: string;
+};
+
 export type GetAppNameAndVersionResponse = {
   appName: string;
   version: string;
@@ -76,6 +89,10 @@ export type LedgerBridge<T extends LedgerBridgeOptions> = {
   deviceSignTypedData(
     params: LedgerSignTypedDataParams,
   ): Promise<LedgerSignTypedDataResponse>;
+
+  deviceSignDelegationAuthorization(
+    params: LedgerSignDelegationAuthorizationParams,
+  ): Promise<LedgerSignDelegationAuthorizationResponse>;
 
   /**
    * Method to retrieve the name and version of the running application on the Ledger device.
--- a/packages/keyring-eth-ledger-bridge/src/ledger-bridge.ts
+++ b/packages/keyring-eth-ledger-bridge/src/ledger-bridge.ts
@@ -32,9 +32,12 @@ export type LedgerSignDelegationAuthorizationParams = {
   nonce: number;
 };
 
-export type LedgerSignDelegationAuthorizationResponse = Awaited<
-  ReturnType<LedgerHwAppEth['signTransaction']>
->;
+// TODO: Replace with 7702 return type
+export type LedgerSignDelegationAuthorizationResponse = Awaited<{
+  s: string;
+  v: string;
+  r: string;
+}>;
 
 export type GetAppNameAndVersionResponse = {
   appName: string;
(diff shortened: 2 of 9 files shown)


packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -720,6 +772,26 @@ export class LedgerKeyring implements Keyring {
...
+    if (typeof recoveryParam === 'number') {
+      return recoveryParam;
+    }
+
+    const value = String(recoveryParam).trim();
+    const withoutPrefix =
+      value.startsWith('0x') || value.startsWith('0X') ? value.slice(2) : value;
+    const radix = /[a-f]/iu.test(withoutPrefix) ? 16 : 10;
+
+    return parseInt(withoutPrefix, radix);
```
ccharly: Actually I think the logic is wrong here.

If we have a number like `0x100` (= 256), then this would be parsed as `100` (radix=10) which is wrong.

IMO we can do without the regex too:

```suggestion
    const value = String(recoveryParam).trim();
    
    if (value.startsWith('0x') || value.startsWith('0X')) {
      return parseInt(value.slice(2), 16);
    }
    
    return parseInt(value, 10);
```

My suggestion is not fully safe either... If the input is `0x` or `0X`, then `parseInt` will return `NaN`. I'm not sure what we should do here? Cause initially we were not even filtering those `NaN` 😅 I guess we could have an `assert(result !== NaN)` too? 🤔

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -593,6 +594,57 @@ export class LedgerKeyring implements Keyring {
...
+        'Ledger: Unknown error while signing EIP-7702 authorization',
+      );
+    }
+
+    const recoveryValue = this.#parseLedgerRecoveryParam(payload.v);
+    const yParity =
+      recoveryValue === 0 || recoveryValue === 1
+        ? recoveryValue
+        : recoveryValue - 27;
+    const signature = `0x${payload.r}${payload.s}${yParity.toString(16).padStart(2, '0')}`;
```
ccharly: Would be nice to have some comments here.

Also, I wouldn't compute the hex within the format string. Maybe we could have a `toSomething(recoveryValue)` that would do the right thing? Including the padding formatting?

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -593,6 +594,57 @@ export class LedgerKeyring implements Keyring {
...
+  async signEip7702Authorization(
+    withAccount: Hex,
+    authorization: [chainId: number, contractAddress: Hex, nonce: number],
+  ): Promise<string> {
+    const [chainId, contractAddress, nonce] = authorization;
+    const hdPath = await this.unlockAccountByAddress(withAccount);
+
+    if (!hdPath) {
+      throw new Error(
+        'Ledger: Unknown error while signing EIP-7702 authorization',
```
gantunesr: Should this reflect that the error is related to the missing `hdPath`?

```suggestion
        'Ledger: Missing hdPath while signing EIP-7702 authorization',
```

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -593,6 +594,57 @@ export class LedgerKeyring implements Keyring {
...
+        hdPath,
+        chainId,
+        contractAddress: remove0x(contractAddress),
+        nonce,
+      });
+    } catch (error: unknown) {
+      handleLedgerTransportError(
+        error,
+        'Ledger: Unknown error while signing EIP-7702 authorization',
+      );
```
gantunesr, line 624: I think we can also remove the `Unknown` here since we have the actual `error` as the first param for the handler

```suggestion
    } catch (error: unknown) {
      handleLedgerTransportError(
        error,
        'Ledger: Error while signing EIP-7702 authorization',
      );
```

packages/keyring-eth-ledger-bridge/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added 7702 methods ([#564](https://github.com/MetaMask/accounts/pull/564))
```
gantunesr:
```suggestion
- Add EIP-7702 authorization signing to the Ledger keyring ([#564](https://github.com/MetaMask/accounts/pull/564))
    - Not yet functional: both iframe and mobile bridges throw until DMK support lands.
    - Fixes hex `v` parsing (`'1b'`) for personal_sign and typed data. 
```

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
gantunesr: For `NaN` I would suggest throwing an error like: `Invalid recovery parameter from device`

gantunesr: And I would also suggest adding test cases that cover the following,

- Hex without prefix: '1b' → 27
- Hex with prefix: '0x1b' → 27
- String: '27' → 27
- Integer: 27 → 27
- Hex prefix: '0x' → throw

packages/keyring-eth-ledger-bridge/src/ledger-bridge.ts
```diff
@@ -25,6 +25,17 @@ export type LedgerSignTypedDataResponse = Awaited<
...
+export type LedgerSignDelegationAuthorizationParams = {
+  hdPath: string;
+  chainId: number;
+  contractAddress: string;
+  nonce: number;
+};
+
+export type LedgerSignDelegationAuthorizationResponse = Awaited<
+  ReturnType<LedgerHwAppEth['signTransaction']>
+>;
```
gantunesr: Is this a placeholder? Should it a delegation signature instead of `signTransaction`?

montelaidev: Its the same return type. It just an object with rsv. We can just add a note to replace it

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
montelaidev: Added additional tests

packages/keyring-eth-ledger-bridge/src/ledger-mobile-bridge.ts
```diff
@@ -102,6 +104,17 @@ export class LedgerMobileBridge implements MobileBridge {
     return this.#getEthApp().signEIP712Message(hdPath, message);
   }
 
+  async deviceSignDelegationAuthorization({
+    hdPath: _hdPath,
+    chainId: _chainId,
+    contractAddress: _contractAddress,
+    nonce: _nonce,
+  }: LedgerSignDelegationAuthorizationParams): Promise<LedgerSignDelegationAuthorizationResponse> {
```
mathieuartu, line 112: Nit: why destructure the arguments if we use none?

packages/keyring-eth-ledger-bridge/src/ledger-iframe-bridge.ts
```diff
@@ -238,6 +240,14 @@ export class LedgerIframeBridge implements LedgerBridge<LedgerIframeBridgeOption
     );
   }
 
+  async deviceSignDelegationAuthorization(
+    _params: LedgerSignDelegationAuthorizationParams,
```
mathieuartu, line 244: Naive Q: why optional `params` here and not elsewhere where `LedgerSignDelegationAuthorizationParams` is used in this PR?

packages/keyring-eth-ledger-bridge/src/ledger-keyring.ts
```diff
@@ -720,6 +774,38 @@ export class LedgerKeyring implements Keyring {
...
+    const value = String(recoveryParam).trim();
+
+    if (value.startsWith('0x') || value.startsWith('0X')) {
+      if (!isStrictHexString(value)) {
+        throw new Error(`Invalid hex recovery parameter: "${recoveryParam}"`);
+      }
+      return hexToNumber(value);
+    }
+
+    const radix = /[a-f]/iu.test(value) ? 16 : 10;
```
mathieuartu, line 801: Nit: would love a comment here explaining what this is

packages/keyring-eth-ledger-bridge/src/ledger-iframe-bridge.ts
montelaidev: The `hw-eth-app` doesn't support it. We will be introducing the DMK bridge that will be using it.

packages/keyring-eth-ledger-bridge/src/ledger-mobile-bridge.ts
montelaidev: The `hw-eth-app` doesn't support it. We will be introducing the DMK bridge that will be using it.

mathieuartu (approved):
LGTM, but might have missed something due to not being 100% up to date on the concept. Maybe wait for a second review before merging? But no issues on my end!

gantunesr (approved):

PR #581, opened 2026-06-30
feat(keyring-snap-bridge): populate v2 SnapKeyring capabilities from the manifest

hmalik88:
- Populate the v2 `SnapKeyring.capabilities` from the Snap manifest (`endowment:keyring`) on `deserialize`, via `SnapController:getSnap`, fulfilling the v2 `Keyring` contract's required `capabilities` field. Falls back to the empty default when the Snap declares none.
- Guard v2 `SnapKeyring` operations (`getAccounts`/`getAccount`/`createAccounts`/`deleteAccount`/`submitRequest`) until `deserialize` has run; they now throw `"SnapKeyring has not been initialized"` instead of operating on an unbound keyring.
- Bump `@metamask/snaps-utils` to `^12.2.1` (`keyring-snap-bridge` + `keyring-internal-snap-client`) for the `capabilities` field on the `endowment:keyring` manifest type.

---

> [!NOTE]
> **Medium Risk**
> Changes snap keyring lifecycle and when account operations are allowed; capabilities are taken from the manifest without local shape validation, so struct drift could cause runtime issues.
> 
> **Overview**
> v2 **`SnapKeyring`** now sets **`capabilities`** during **`deserialize`** by reading the Snap manifest’s **`endowment:keyring`** entry via **`SnapController:getSnap`**, with **`EMPTY_CAPABILITIES`** as the typed default when the snap is missing or declares none (including clearing stale values on re-hydrate).
> 
> **`getAccounts`**, **`getAccount`**, **`createAccounts`**, **`deleteAccount`**, and **`submitRequest`** are blocked until **`deserialize`** finishes, throwing *SnapKeyring has not been initialized* instead of running on an unbound keyring.
> 
> **`@metamask/snaps-utils`** is bumped to **`^12.2.1`** in **`keyring-snap-bridge`** and **`keyring-internal-snap-client`** for manifest typing around **`capabilities`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 39a8b9423f98dfd6024e731ac7bbc4c259704902. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/keyring-snap-bridge/src/v2/SnapKeyring.test.ts
+++ b/packages/keyring-snap-bridge/src/v2/SnapKeyring.test.ts
@@ -158,6 +158,34 @@ describe('SnapKeyring', () => {
       const keyring = await makeKeyringWithSnap(undefined);
       expect(keyring.capabilities).toStrictEqual({ scopes: [] });
     });
+
+    it('clears stale capabilities on a later deserialize without capabilities', async () => {
+      const snapWith = (keyringPermission: unknown): unknown => ({
+        manifest: {
+          initialPermissions: { 'endowment:keyring': keyringPermission },
+        },
+      });
+      const getSnap = jest
+        .fn()
+        .mockReturnValueOnce(snapWith({ capabilities: manifestCapabilities }))
+        .mockReturnValueOnce(snapWith({ allowedOrigins: [] }));
+      const messenger = {
+        call: jest.fn((action: string) =>
+          action === 'SnapController:getSnap' ? getSnap() : undefined,
+        ),
+        publish: jest.fn(),
+      } as unknown as SnapKeyringMessenger;
+      const keyring = new SnapKeyring({
+        messenger,
+        callbacks: makeMockCallbacks(),
+      });
+
+      await keyring.deserialize({ snapId: SNAP_ID, accounts: {} });
+      expect(keyring.capabilities).toStrictEqual(manifestCapabilities);
+
+      await keyring.deserialize({ snapId: SNAP_ID, accounts: {} });
+      expect(keyring.capabilities).toStrictEqual({ scopes: [] });
+    });
   });
 
   describe('initialization guard', () => {
--- a/packages/keyring-snap-bridge/src/v2/SnapKeyring.test.ts
+++ b/packages/keyring-snap-bridge/src/v2/SnapKeyring.test.ts
@@ -160,6 +160,56 @@ describe('SnapKeyring', () => {
     });
   });
 
+  describe('initialization guard', () => {
+    const ERROR = 'SnapKeyring has not been initialized';
+
+    /**
+     * Build a `SnapKeyring` WITHOUT calling `deserialize`.
+     *
+     * @returns The uninitialized keyring.
+     */
+    function makeUninitializedKeyring(): SnapKeyring {
+      const messenger = {
+        call: jest.fn(),
+        publish: jest.fn(),
+      } as unknown as SnapKeyringMessenger;
+      return new SnapKeyring({ messenger, callbacks: makeMockCallbacks() });
+    }
+
+    it('rejects getAccounts before deserialize is called', async () => {
+      await expect(makeUninitializedKeyring().getAccounts()).rejects.toThrow(
+        ERROR,
+      );
+    });
+
+    it('rejects getAccount before deserialize is called', async () => {
+      await expect(
+        makeUninitializedKeyring().getAccount(account1.id),
+      ).rejects.toThrow(ERROR);
+    });
+
+    it('rejects createAccounts before deserialize is called', async () => {
+      await expect(
+        makeUninitializedKeyring().createAccounts({
+          type: 'bip44:derive-index',
+          entropySource: 'mock-entropy-source',
+          groupIndex: 0,
+        } as CreateAccountOptions),
+      ).rejects.toThrow(ERROR);
+    });
+
+    it('rejects deleteAccount before deserialize is called', async () => {
+      await expect(
+        makeUninitializedKeyring().deleteAccount(account1.id),
+      ).rejects.toThrow(ERROR);
+    });
+
+    it('allows operations after deserialize', async () => {
+      const { keyring } = await makeKeyring();
+      expect(await keyring.getAccounts()).toStrictEqual([]);
+    });
+  });
+
   describe('snapId', () => {
     it('returns the snap ID set during deserialize', async () => {
       const { keyring } = await makeKeyring();
--- a/packages/keyring-snap-bridge/src/v2/SnapKeyring.test.ts
+++ b/packages/keyring-snap-bridge/src/v2/SnapKeyring.test.ts
@@ -101,6 +101,65 @@ describe('SnapKeyring', () => {
     });
   });
 
+  describe('capabilities', () => {
+    const manifestCapabilities = {
+      scopes: ['solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp'],
+      bip44: { deriveIndex: true, deriveIndexRange: true, discover: true },
+    };
+
+    /**
+     * Build and deserialize a keyring whose messenger returns a snap with the
+     * given `endowment:keyring` permission value from `SnapController:getSnap`.
+     *
+     * @param keyringPermission - The `endowment:keyring` initial-permission value
+     * (or `undefined` to simulate a snap that is not found).
+     * @returns The deserialized keyring.
+     */
+    async function makeKeyringWithSnap(
+      keyringPermission: unknown,
+    ): Promise<SnapKeyring> {
+      const messenger = {
+        call: jest.fn((action: string) =>
+          action === 'SnapController:getSnap' && keyringPermission !== undefined
+            ? {
+                manifest: {
+                  initialPermissions: {
+                    'endowment:keyring': keyringPermission,
+                  },
+                },
+              }
+            : undefined,
+        ),
+        publish: jest.fn(),
+      } as unknown as SnapKeyringMessenger;
+      const keyring = new SnapKeyring({
+        messenger,
+        callbacks: makeMockCallbacks(),
+      });
+      await keyring.deserialize({ snapId: SNAP_ID, accounts: {} });
+      return keyring;
+    }
+
+    it('populates capabilities from the snap manifest on deserialize', async () => {
+      const keyring = await makeKeyringWithSnap({
+        capabilities: manifestCapabilities,
+      });
+      expect(keyring.capabilities).toStrictEqual(manifestCapabilities);
+    });
+
+    it('keeps the empty default when the snap declares no capabilities', async () => {
+      const keyring = await makeKeyringWithSnap({
+        allowedOrigins: ['https://portfolio.metamask.io'],
+      });
+      expect(keyring.capabilities).toStrictEqual({ scopes: [] });
+    });
+
+    it('keeps the empty default when the snap cannot be found', async () => {
+      const keyring = await makeKeyringWithSnap(undefined);
+      expect(keyring.capabilities).toStrictEqual({ scopes: [] });
+    });
+  });
+
   describe('snapId', () => {
     it('returns the snap ID set during deserialize', async () => {
       const { keyring } = await makeKeyring();
--- a/packages/keyring-snap-bridge/src/v2/SnapKeyring.test.ts
+++ b/packages/keyring-snap-bridge/src/v2/SnapKeyring.test.ts
@@ -10,7 +10,7 @@ import type { SnapId } from '@metamask/snaps-sdk';
 
 import type { SnapKeyringMessenger } from '../SnapKeyringMessenger';
 import type { SnapKeyringCallbacks } from './SnapKeyring';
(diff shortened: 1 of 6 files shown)


packages/keyring-snap-bridge/src/v2/SnapKeyring.ts
```diff
@@ -487,6 +498,12 @@ export class SnapKeyring extends SnapKeyringV1 implements Keyring {
     // mismatch to prevent swapping a keyring's identity via deserialize).
     this.bindSnapId(state.snapId as SnapId);
 
+    const capabilities = this.#resolveKeyringCapabilities();
+
+    if (capabilities) {
+      this.capabilities = capabilities;
+    }
```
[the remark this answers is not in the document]
hmalik88: [7282326](https://github.com/MetaMask/accounts/pull/581/commits/7282326c9a09dcfcb698f3066de675161206877b)

packages/keyring-snap-bridge/src/v2/SnapKeyring.ts
```diff
@@ -508,9 +525,25 @@ export class SnapKeyring extends SnapKeyringV1 implements Keyring {
...
+      throw new Error(
+        'SnapKeyring has not been initialized: call deserialize() first',
+      );
+    }
+  }
+
+  #resolveKeyringCapabilities(): KeyringCapabilities | undefined {
+    const snap = this.messenger.call('SnapController:getSnap', this.snapId);
+    return snap?.manifest.initialPermissions['endowment:keyring']
+      ?.capabilities as KeyringCapabilities | undefined;
```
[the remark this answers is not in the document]
hmalik88: optional chaining short circuits the whole chain

packages/keyring-snap-bridge/src/v2/SnapKeyring.ts
```diff
@@ -487,6 +498,13 @@ export class SnapKeyring extends SnapKeyringV1 implements Keyring {
     // mismatch to prevent swapping a keyring's identity via deserialize).
     this.bindSnapId(state.snapId as SnapId);
 
+    // Refresh capabilities from the snap manifest on every deserialize, falling
+    // back to the empty default so a re-hydrate clears any previously-loaded
+    // capabilities when the snap no longer declares them.
+    this.capabilities =
+      this.#resolveKeyringCapabilities() ??
+      ({ scopes: [] } as unknown as KeyringCapabilities);
```
ccharly: Would be nice to have an `EMPTY_CAPABILITIES` constant for this.

Also, we shouldn't need the `as unknown` cast for this 🤔 I think `{ scopes: [] }` is the minimum we need.

Using the constant + removing the `as unknown` would also ensure us that we would not skip any future non-optional fields for `capabilities` whenever we add new ones!

packages/keyring-snap-bridge/src/v2/SnapKeyring.ts
```diff
@@ -508,9 +526,25 @@ export class SnapKeyring extends SnapKeyringV1 implements Keyring {
...
+
+  #assertInitialized(): void {
+    if (!this.#initialized) {
+      throw new Error(
+        'SnapKeyring has not been initialized: call deserialize() first',
+      );
+    }
+  }
+
+  #resolveKeyringCapabilities(): KeyringCapabilities | undefined {
```
ccharly, line 561: I know those are private, but could you add a small jsdoc for those too?

hmalik88: [21f8a46](https://github.com/MetaMask/accounts/pull/581/commits/21f8a468c8c753e66c7d7bbc78f8df1674e0ffa6)

packages/keyring-snap-bridge/src/v2/SnapKeyring.ts
hmalik88: [9ac9ce9](https://github.com/MetaMask/accounts/pull/581/commits/9ac9ce9f846850c7063bca0f0427bd449a09afaf)

packages/keyring-snap-bridge/src/v2/SnapKeyring.ts
```diff
@@ -27,6 +27,13 @@ import type {
 } from '../SnapKeyringV1';
 import { equalsIgnoreCase } from '../util';
 
+/**
+ * Default, empty capabilities used until the snap manifest is read on
+ * `deserialize`. Typed (no cast) so that adding a new required
+ * `KeyringCapabilities` field surfaces here as a compile error.
+ */
+const EMPTY_CAPABILITIES: KeyringCapabilities = { scopes: [] };
```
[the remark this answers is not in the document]
hmalik88: [c11309e](https://github.com/MetaMask/accounts/pull/581/commits/c11309ea634ea6f4111d6f960d94df7fa481b0fb)

packages/keyring-snap-bridge/src/v2/SnapKeyring.ts
```diff
@@ -508,9 +530,35 @@ export class SnapKeyring extends SnapKeyringV1 implements Keyring {
...
+
+  /**
+   * Resolve the keyring capabilities from the snap manifest.
+   *
+   * @returns The keyring capabilities, or `undefined` if the snap manifest does not declare any capabilities.
+   */
+  #resolveKeyringCapabilities(): KeyringCapabilities | undefined {
+    const snap = this.messenger.call('SnapController:getSnap', this.snapId);
+    return snap?.manifest.initialPermissions['endowment:keyring']
+      ?.capabilities as KeyringCapabilities | undefined;
```
ccharly, line 570: I know we should trust those data, but since we're casting them here, there's a change that schema might diverge a bit and thus, potentially causing serious runtime issues (since we'll rely a lot on those capabilities soon).

WDYT about validating them against `KeyringCapabilitiesStruct`? If anything goes wrong, we default to `EMPTY_CAPABILITIIES` and log something?

Or maybe we let the error bubble up 🤔 this would be considered like an "unsupported keyring" (all keyrings that throw during init/deserialize) will end up there, so that might be a good thing actually.

Any opinion?

hmalik88: Yeah that makes sense, the struct is good defensive behavior

hmalik88: Actually on second thought, there's manifest validation done on the snaps side already. So this feels like it would be unnecessary. The `KeyringCapabilitiesStruct` in the snap repo was written to mirror what we have in the accounts repo: https://github.com/MetaMask/snaps/blob/main/packages/snaps-utils/src/keyring.ts#L46

hmalik88: [6201c52](https://github.com/MetaMask/accounts/pull/581/commits/6201c52051a3964e63ec79c3460cdee69cf6deb7)

packages/keyring-snap-bridge/src/v2/SnapKeyring.ts
```diff
@@ -508,9 +530,38 @@ export class SnapKeyring extends SnapKeyringV1 implements Keyring {
...
+  /**
+   * Resolve the keyring capabilities from the snap manifest.
+   *
+   * @returns The keyring capabilities, or `undefined` if the snap manifest does not declare any capabilities.
+   */
+  #resolveKeyringCapabilities(): KeyringCapabilities | undefined {
+    const snap = this.messenger.call('SnapController:getSnap', this.snapId);
+    // we are not validating the shape of the capabilities here, because there is
+    // manifest validation done already on the snaps side, the snaps repo maintains
+    // a copy of the KeyringCapabilitiesStruct
```
ccharly:
```suggestion
    // READ THIS CAREFULLY:
    // We are not validating the shape of the capabilities here, because there is
    // manifest validation done already on the snaps side, the snaps repo maintains
    // a copy of the `KeyringCapabilitiesStruct`.
    // We must ensure that both structs are always in-sync, otherwise the type-cast
    // could cause runtime issues!
```

packages/keyring-snap-bridge/src/v2/SnapKeyring.ts
ccharly: Hold on, after checking the actual type, the `initialPermissions` are well-typed already.

Which means, if both schemas (ours and theirs) are in-sync. we should not need any type-cast.

For some reason, this fails right now, so they maybe already diverged, which means we would need to update them again (first on the `snaps` side, then update the relevant packages to be able to remove the type-cast).

packages/keyring-snap-bridge/src/v2/SnapKeyring.ts
ccharly: Might not be relevant if we do this: https://github.com/MetaMask/accounts/pull/581#discussion_r3505593037

packages/keyring-snap-bridge/src/v2/SnapKeyring.ts
ccharly: Ok, let's go with that for now. I'll follow-up with a type fix soon after. I need to release the `snaps` monorepo first for:
- https://github.com/MetaMask/snaps/pull/4057

packages/keyring-snap-bridge/src/v2/SnapKeyring.ts
hmalik88: [879b5f2](https://github.com/MetaMask/accounts/pull/581/commits/879b5f2efc6a3dc4a16e117c9982c1cb55688978)

packages/keyring-snap-bridge/src/v2/SnapKeyring.test.ts
```diff
@@ -101,6 +101,143 @@ describe('SnapKeyring', () => {
...
+    it('keeps the empty default when the snap declares no capabilities', async () => {
+      const keyring = await makeKeyringWithSnap({
+        allowedOrigins: ['https://portfolio.metamask.io'],
+      });
+      expect(keyring.capabilities).toStrictEqual({ scopes: [] });
+    });
+
+    it('keeps the empty default when the snap cannot be found', async () => {
+      const keyring = await makeKeyringWithSnap(undefined);
+      expect(keyring.capabilities).toStrictEqual({ scopes: [] });
```
ccharly: Could you re-use `EMPTY_CAPABILITIES` here? We'll need to `export` it too, but I think that's fine, since we might want to use this if a Snap keyring does support v2 or not (outside of that package I mean)

packages/keyring-snap-bridge/src/v2/SnapKeyring.test.ts
```diff
@@ -101,6 +101,143 @@ describe('SnapKeyring', () => {
...
+      const keyring = new SnapKeyring({
+        messenger,
+        callbacks: makeMockCallbacks(),
+      });
+
+      await keyring.deserialize({ snapId: SNAP_ID, accounts: {} });
+      expect(keyring.capabilities).toStrictEqual(manifestCapabilities);
+
+      await keyring.deserialize({ snapId: SNAP_ID, accounts: {} });
+      expect(keyring.capabilities).toStrictEqual({ scopes: [] });
```
ccharly: Same here?

ccharly: Could we re-write this like:
```suggestion
      const mockGetSnap = jest.fn();
      const messenger = {
        call: jest.fn((action: string) =>
          action === 'SnapController:getSnap' ? getSnap() : undefined,
        ),
        publish: jest.fn(),
      } as unknown as SnapKeyringMessenger;
      const keyring = new SnapKeyring({
        messenger,
        callbacks: makeMockCallbacks(),
      });

      mockGetSnap.mockReturnValueOnce(snapWith({ capabilities: manifestCapabilities }))
      await keyring.deserialize({ snapId: SNAP_ID, accounts: {} });
      expect(keyring.capabilities).toStrictEqual(manifestCapabilities);

      // Manifest won't yield any capabilities this time, so we end up
      // clearing the old ones and fallback to the `EMPTY_CAPABILITIES`.
      mockGetSnap.mockReturnValueOnce(snapWith({ allowedOrigins: [] }));
      await keyring.deserialize({ snapId: SNAP_ID, accounts: {} });
      expect(keyring.capabilities).toStrictEqual(EMPTY_CAPABILITIES);
```

hmalik88: [39a8b94](https://github.com/MetaMask/accounts/pull/581/commits/39a8b9423f98dfd6024e731ac7bbc4c259704902)

hmalik88: [39a8b94](https://github.com/MetaMask/accounts/pull/581/commits/39a8b9423f98dfd6024e731ac7bbc4c259704902)

packages/keyring-snap-bridge/src/v2/SnapKeyring.test.ts
hmalik88: [39a8b94](https://github.com/MetaMask/accounts/pull/581/commits/39a8b9423f98dfd6024e731ac7bbc4c259704902)

ccharly (approved):

PR #584, opened 2026-07-05
refactor(keyring-snap-bridge)!: remove `SnapKeyringV1` inheritance

ccharly:
Removing inheritance so we the remaining v1 usage are explicit, since `SnapKeyring` (v2) was implicitly exposing v1 methods, it made it pretty difficult to sunset v1 completely.

---

> [!NOTE]
> **High Risk**
> Breaking public API and behavior split across account creation, signing, and snap messages in security-sensitive keyring code; incorrect v1/v2 routing could break snaps or block v2-only accounts.
> 
> **Overview**
> **Breaking:** v2 `SnapKeyring` no longer extends `SnapKeyringV1`. V1 behavior is exposed only via an optional `.v1` getter (set when the snap manifest has no `endowment:keyring` capabilities; `undefined` for v2-only snaps).
> 
> The per-snap v2 class now owns its registry and v2 `KeyringInternalSnapClient`, and on `deserialize` may compose a `SnapKeyringV1` that shares the same registry. `createAccounts` / `submitRequest` route to v1 or the v2 client depending on snap version (including the v1 `{ options }` wrapper vs flat v2 options).
> 
> The top-level `SnapKeyring` coordinator routes legacy v1 entry points (messages, signing, user ops, etc.) through `getKeyringV1For` and throws clear errors when the target snap is v2-only. `setSelectedAccounts` uses `keyring.v1?.` so v2 snaps are skipped.
> 
> `SnapKeyringV1` gains optional shared `registry`, public `bindSnapId`, and `createAccounts` for v1 RPC. Changelog and tests document the API shift and v1/v2 routing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 693e6f33923b2eb38e86257fbcd7042e19d8544b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/keyring-snap-bridge/src/SnapKeyring.ts
+++ b/packages/keyring-snap-bridge/src/SnapKeyring.ts
@@ -799,6 +799,8 @@ export class SnapKeyring {
       bySnap.set(snapId, snapAccounts);
     }
 
+    // v1 snaps receive a keyring_setSelectedAccounts notification; v2 snaps
+    // have no equivalent concept in the protocol so they are intentionally skipped.
     await Promise.all(
       [...this.#snapKeyrings.entries()].map(async ([snapId, keyring]) =>
         keyring.v1?.setSelectedAccounts(
--- a/packages/keyring-snap-bridge/src/SnapKeyring.ts
+++ b/packages/keyring-snap-bridge/src/SnapKeyring.ts
@@ -88,9 +88,10 @@ export class SnapKeyring {
   readonly #messenger: SnapKeyringMessenger;
 
   /**
-   * Per-snap keyring instances. Each `SnapKeyringV2` (which extends
-   * `SnapKeyringV1`) owns a single `KeyringAccountRegistry` and handles
-   * both the event-driven v1 flow and the `KeyringV2` batch interface.
+   * Per-snap keyring instances. Each `SnapKeyringV2` owns the
+   * `KeyringAccountRegistry` and handles the `KeyringV2` batch interface.
+   * For v1 snaps, `SnapKeyringV2.v1` holds a `SnapKeyringV1` instance
+   * (with a shared registry reference) for the event-driven flow.
    */
   readonly #snapKeyrings: Map<SnapId, SnapKeyringV2>;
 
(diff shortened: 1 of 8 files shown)


packages/keyring-snap-bridge/src/SnapKeyring.ts
```diff
@@ -344,7 +363,10 @@ export class SnapKeyring {
     }
 
     try {
-      return await keyring.handleKeyringSnapMessage(message);
+      return await getKeyringV1For(
```
hmalik88, line 366: This is the only thing i'm not entirely clear on. This would reject any v2 messages? Where do those get routed?

ccharly: Ah yeah, good question 😄

Basically, we want to drop the support of most messages here with v2. Today, the `SnapAccountService` handles the routing, it used to be the legacy `SnapKeyring` (this file here) that were handling all of this, but since the introduction of the service, it now goes through the service first, and might be dispatched to other controllers directly (e.g for assets):
- https://github.com/MetaMask/core/blob/359eb4a47a5bf7e84d94d145c4971fdf5ae714e1/packages/snap-account-service/src/SnapAccountService.ts#L577-L579

If we hit a `setSelectedAccounts`, then the service also handles that:
- https://github.com/MetaMask/core/blob/359eb4a47a5bf7e84d94d145c4971fdf5ae714e1/packages/snap-account-service/src/SnapAccountService.ts#L582-L599

Though, we still delegate some call to the keyring itself for the old v1 flows (for `:account{Created,Updated,Removed}` mainly):
- https://github.com/MetaMask/core/blob/359eb4a47a5bf7e84d94d145c4971fdf5ae714e1/packages/snap-account-service/src/SnapAccountService.ts#L615-L641

So yes, you're right, there are no "v2 messages" being routed here anymore. Because:
- this class is the old legacy `SnapKeyring`, we mainly keep it to initiate the migration from the old state (1:N) to new the v2 layout (1:1).
- most routing goes through the service now
- the remaining v1 flows messages are still being routed to a proper `keyring.v1` instance (we need that for Snaps that did not implement v2 yet)

hmalik88: ok perfect, thanks for the explanation!

hmalik88 (approved):
Approving with the caveat that we don't merge until we test integration against the various preview builds we have up (sol, core, this)

ccharly:
Was able to create account and the new `.v1` is well defined when testing with the extension. Merging now so we can release and update `core` and the Solana Snap!

PR #590, opened 2026-07-08
feat(account-api): add `PrivateKeyAccount` type helper

ccharly:
Adding a `PrivateKeyAccount<Account>` type helper, similar to `Bip44Account<Account>`. This new one automatically forces the `options.entropy.type` to be `'private-key'`.

---

> [!NOTE]
> **Low Risk**
> Additive API and a small enum-based change to BIP-44 detection; behavior should be equivalent if enum values match prior string literals.
> 
> **Overview**
> Adds **`PrivateKeyAccount<Account>`** alongside the existing BIP-44 pattern: a typed account shape with `options.entropy` narrowed to private-key entropy, plus **`isPrivateKeyAccount`** and **`assertIsPrivateKeyAccount`** guards keyed off `KeyringAccountEntropyTypeOption.PrivateKey`. Unit tests cover positive and negative entropy/type cases.
> 
> **`isBip44Account`** now compares entropy type using **`KeyringAccountEntropyTypeOption.Mnemonic`** instead of the string literal `'mnemonic'`, matching the new private-key checks and `@metamask/keyring-api` enums.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 554b30a8079c1d195b8d1158e013ae3068d5c8c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/account-api/src/api/private-key.ts
+++ b/packages/account-api/src/api/private-key.ts
@@ -4,7 +4,7 @@ import type {
 } from '@metamask/keyring-api';
 
 /**
- * Private key compatible account type.
+ * Private key account type.
  */
 export type PrivateKeyAccount<Account extends KeyringAccount> = Account & {
   // We force the option type for those accounts. (That's how we identify
@@ -15,10 +15,10 @@ export type PrivateKeyAccount<Account extends KeyringAccount> = Account & {
 };
 
 /**
- * Checks if an account is private key compatible.
+ * Checks if an account is a private key account.
  *
  * @param account - The account to be tested.
- * @returns True if the account is private key compatible.
+ * @returns True if the account is a private key account.
  */
 export function isPrivateKeyAccount<Account extends KeyringAccount>(
   account: Account,
@@ -29,10 +29,10 @@ export function isPrivateKeyAccount<Account extends KeyringAccount>(
 }
 
 /**
- * Asserts a keyring account is private keycompatible.
+ * Asserts a keyring account is a private key account.
  *
  * @param account - Keyring account to check.
- * @throws If the keyring account is not compatible.
+ * @throws If the keyring account is not a private key account.
  */
 export function assertIsPrivateKeyAccount<Account extends KeyringAccount>(
   account: Account,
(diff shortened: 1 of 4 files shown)


packages/account-api/src/api/private-key.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Checks if an account is a private key account.
+ *
+ * @param account - The account to be tested.
+ * @returns True if the account is a private key account.
+ */
+export function isPrivateKeyAccount<Account extends KeyringAccount>(
+  account: Account,
+): account is PrivateKeyAccount<Account> {
+  // To be private key compatible, we just check for the entropy type (the
+  // the `entropy` shape will be inferred automatically).
```
gantunesr: nit
```suggestion
  // To be private key compatible, we just check for the entropy type
  // (the `entropy` shape will be inferred automatically).
```

packages/account-api/src/api/private-key.ts
```diff
@@ -0,0 +1,43 @@
...
+  return account.options.entropy?.type === 'private-key';
```
gantunesr: we don't have an enum to replace the `'private-key'` or do we? I see there is `KeyringAccountEntropyTypeOption` and `KeyringType` but I'm no sure if they fit this context

ccharly: We stopped using enums for the v2 API, that being said, we could introduce (if we don't have one) a `const` object for this maybe. That would result in the same than using an enum (minus the other downside of using real enums :trollface:)

Let me check that

gantunesr: Oh right, `enums` and their breaking changes. Not a blocker imo, just a question. Go with your instinct

gantunesr (dismissed):

packages/account-api/src/api/private-key.ts
ccharly: We actually have an `enum` for it 🤦 so "We stopped using enums for the v2 API" is actually wrong 😅

Though, since we use literals for the entropy object, I'll have to use:
```ts
return account.options.entropy?.type === `${KeyringAccountEntropyTypeOption.PrivateKey}`;
```

Which is ok IMO

ccharly: https://github.com/MetaMask/accounts/pull/590/commits/554b30a8079c1d195b8d1158e013ae3068d5c8c3

gantunesr (approved):

PR #593, opened 2026-07-09
ci: reduce CI runtime and drop EOL Node.js versions

hmalik88:
## Summary

- Remove the `prepare` job and its `needs: prepare` gate from all downstream jobs
- Consolidate `validate-changelog` from 16 matrix jobs to one
- Update `actions/checkout` from v4 to v5
- Drop EOL Node.js versions (18, 20, 24) and test on 22.x only; set minimum engine to `>=22`
- Remove duplicate `yarn constraints` call from the `lint` script

## Why

**`prepare` was adding ~52s to every run for no reason.** Every job in
`build-lint-test.yml` had `needs: prepare`, meaning nothing could start
until prepare finished. The only thing prepare produced that any job
actually consumed was a workspace package names list, used to fan out
`validate-changelog` into a per-package matrix. `lint`, `build`, and
`test` had no real dependency on it — they were blocked purely as a side
effect of the matrix design.

**The `validate-changelog` matrix was 16 jobs for one command.**
`changelog:validate` has no cross-package dependencies, so 16 separate
checkout+setup cycles (each ~35–40s of overhead) for a script that takes
a few seconds is pure waste. `yarn workspaces foreach --all --no-private
--parallel run changelog:validate` does the same work in one job.
Collapsing the matrix also eliminates the need for `prepare`'s output,
which is what makes removing `prepare` possible.

**`actions/checkout@v4` targets Node.js 20**, which GitHub Actions runners
now force to run on Node.js 24, generating a deprecation annotation on
every run.

**Node.js 18 and 20 are both past end-of-life** (April 2025 and April
2026 respectively) and no longer receive security patches. Testing against
them provides no signal and wastes runner slots. Node 22 is Active LTS
until April 2027.

**`yarn constraints` was running twice per CI push.** The `lint` script
called it explicitly, and the dedicated `Lint (constraints)` CI job also
ran it as a separate matrix entry. Since they run in parallel there was no
wall-clock cost, but it was redundant work. Removed the call from the
`lint` script so it only runs in the dedicated job.

## Impact

| | Before | After |
|---|---|---|
| Critical path (measured) | ~414s avg | ~219s |
| `validate-changelog` runner slots | 16 | 1 |
| Test matrix entries | 3 (18, 20, 22) | 1 (22) |
| `yarn constraints` runs per push | 2 | 1 |
| Deprecation warnings | yes | no |

About a 30% improvement, which will probably be even better when caches are warm. This should also make the range of CI times more consistent.

## Test plan

- [x] All jobs pass on this PR's CI run
- [x] Wall-clock time is materially shorter than recent runs on `main`
- [x] `validate-changelog` catches a bad changelog entry (manual spot-check)

---

> [!NOTE]
> **Medium Risk**
> Raising minimum Node to 22 is a breaking change for downstream consumers on 18/20; CI-only edits are low risk but affect release gates for all packages.
> 
> **Overview**
> **CI is restructured to start jobs immediately and run less redundant work.** The `prepare` job and all `needs: prepare` edges are removed. Changelog validation is a single job using `yarn workspaces foreach --all --no-private --parallel run changelog:validate` instead of a per-package matrix fed by `prepare`.
> 
> **Node and lint behavior change across the repo.** Root and workspace `engines.node` move from `^18.18 || >=20` to `>=22`, enforced in `yarn.config.cjs`, with matching Unreleased changelog notes. Tests run only on Node 22.x; build and lint matrices add 22.x and 24.x with explicit `node-version` on checkout setup. CI lint uses a new `lint:ci` script (eslint, misc, dependencies, readme—no `yarn constraints`); the dedicated `constraints` matrix entry still runs constraints. `actions/checkout` is bumped to v5 in `main.yml`.
> 
> A stale ESLint suppression for `packages/keyring-eth-hd/src/hd-keyring.test.ts` is removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6edb055563d95c76fc10177bb166580b39a02b07. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


package.json
```diff
@@ -19,7 +19,7 @@
     "build:types": "tsc --build tsconfig.build.json --verbose",
     "create-release-branch": "create-release-branch --formatter oxfmt",
     "foreach": "yarn workspaces foreach --all --parallel --verbose --exclude '@metamask/accounts-monorepo' run",
-    "lint": "yarn lint:eslint && yarn lint:misc --check && yarn constraints && yarn lint:dependencies && yarn lint:readme",
```
ccharly, line 22: Why are we getting rid of `yarn constraints` here? We should keep it

.github/workflows/build-lint-test.yml
ccharly, line 1: Maybe we could also do the same as this:
- https://github.com/MetaMask/core/pull/8500

Looks like node 24 is much faster for some of those jobs

package.json
hmalik88: Hmm good catch. I removed `yarn constraints` because it was being ran twice per CI push; once inside `yarn lint` (via the Lint (lint) job) and again in the dedicated Lint (constraints) job. Rather than dropping it entirely, I've added a lint:ci script that CI uses instead, identical to `yarn lint` but without the constraints call.

[1465697](https://github.com/MetaMask/accounts/pull/593/commits/1465697259db259d1ae9d09f2394b4eaa5ca19c7)

.github/workflows/build-lint-test.yml
hmalik88: [e0c7726](https://github.com/MetaMask/accounts/pull/593/commits/e0c77261fafcdd7c7dbd468853d660c6eac6cf6e)

packages/keyring-utils/CHANGELOG.md
```diff
@@ -7,10 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Changed
+
+- Drop support for Node.js v18 and v20; minimum version is now v22 ([#593](https://github.com/MetaMask/accounts/pull/593))
+
 ## [4.0.0]
 
 ### Changed
 
+- Drop support for Node.js v18 and v20; minimum version is now v22 ([#593](https://github.com/MetaMask/accounts/pull/593))
```
ccharly:
```suggestion
```

packages/keyring-snap-sdk/CHANGELOG.md
```diff
@@ -7,10 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Changed
+
+- Drop support for Node.js v18 and v20; minimum version is now v22 ([#593](https://github.com/MetaMask/accounts/pull/593))
+
 ## [9.2.1]
 
 ### Changed
 
+- Drop support for Node.js v18 and v20; minimum version is now v22 ([#593](https://github.com/MetaMask/accounts/pull/593))
```
ccharly:
```suggestion
```

packages/keyring-snap-bridge/CHANGELOG.md
```diff
@@ -7,10 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Changed
+
+- Drop support for Node.js v18 and v20; minimum version is now v22 ([#593](https://github.com/MetaMask/accounts/pull/593))
+
 ## [23.0.2]
 
 ### Changed
 
+- Drop support for Node.js v18 and v20; minimum version is now v22 ([#593](https://github.com/MetaMask/accounts/pull/593))
```
ccharly:
```suggestion
```

packages/keyring-sdk/CHANGELOG.md
```diff
@@ -20,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Drop support for Node.js v18 and v20; minimum version is now v22 ([#593](https://github.com/MetaMask/accounts/pull/593))
```
ccharly:
```suggestion
```

packages/keyring-api/CHANGELOG.md
```diff
@@ -24,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Drop support for Node.js v18 and v20; minimum version is now v22 ([#593](https://github.com/MetaMask/accounts/pull/593))
```
ccharly:
```suggestion
```

ccharly (approved):

PR #601, opened 2026-07-22
feat(keyring-api): add optional `Balance.metadata`

ccharly:
Adding a new `metadata` field for `BalanceStruct`. This way, asset balances can now provide more "specific" information alongside the current balance.

There is not defined shape, we leave it for the consumer and producer to be in-sync regarding the content of it.

---

> [!NOTE]
> **Low Risk**
> Backward-compatible optional API/schema extension with tests only; no runtime or security-sensitive logic changes in this diff.
> 
> **Overview**
> Adds an optional **`metadata`** field on **`BalanceStruct`**: a string-keyed map of JSON-compatible values with no prescribed shape, so producers and consumers agree on meaning out of band.
> 
> **`FungibleAssetAmountStruct`** now spreads **`BalanceStruct.schema`** instead of inlining `unit` and `amount`, so fungible asset amounts pick up **`metadata`** automatically (including on balance-related notifications such as **`notify:accountBalancesUpdated`** per the changelog). Validation rejects non-object metadata, primitives, explicit `undefined` on the key, and nested values that are not JSON-safe.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd8292c58f04fe580283c1069073cb8c36fa7cd6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/keyring-api/src/api/asset.ts
```diff
@@ -1,25 +1,19 @@
...
-  /**
-   * Asset unit.
-   */
-  unit: string(),
-
-  /**
-   * Asset amount.
-   */
-  amount: StringNumberStruct,
+  ...BalanceStruct.schema,
```
ccharly, line 16: Just re-inject this schema to avoid drifting. (I almost missed it initially, but Cursor reported it, so I think this refactor makes sense to avoid that in the future)

danroc (approved):

PR #613, opened 2026-07-31
feat: reuse paging-derived accounts to avoid redundant device I/O at unlock

gantunesr:
Summary of the changes:

- New private `#pagingCache: Map<string, { hdPath, address, publicKey, chainCode? }>` keyed by derivation path.
- `#getAccountsBIP44` now calls `bridge.getPublicKey` directly instead of `this.unlock(path)`, so it can capture the full payload and stash `{ hdPath, address, publicKey: payload.publicKey, chainCode: payload.chainCode }` into the cache. Error handling and the `hdk` update are preserved.
- `addAccounts` checks `#pagingCache.get(path)` first; on a hit it reuses `cached.address` and skips `unlock(path)` → `bridge.getPublicKey`. On a miss it falls back to `await this.unlock(path)`. Note: The non-Ledger Live branch is unchanged.
- Cache is invalidated on `forgetDevice`, `setHdPath` (when the path actually changes), `destroy`, and defensively at the start of `addAccounts` when the device is disconnected (`!this.isConnected()`).
- New public `invalidatePagingCache()` method for external disconnect handlers, backed by private `#invalidatePagingCache()`.

## Examples

N/A


gantunesr:
@metamaskbot publish-preview

PR #618, opened 2026-08-06
feat(keyring-snap-bridge): add `isSnapKeyringV1Adapter`

ccharly:
Adding this helper since I have faced some issue with the error "Keyring not found" on the extension.

That's probably because of the same issue we had with the `KeyringControllerError` a while ago that was also using `instanceof`, if multiple majors are in the dep-tree, the classes could be seen as different and make the test fail!

---

> [!NOTE]
> **Low Risk**
> Read-only identification helper with no changes to keyring behavior or account operations.
> 
> **Overview**
> Adds **`isSnapKeyringV1Adapter`**, a type guard so callers can recognize a Snap v1 adapter without **`instanceof`**, which can fail when multiple copies of the same class are loaded (e.g. extension dependency trees).
> 
> The check is duck-typed: **`type === KeyringType.Snap`** and **`unwrap`** is a function. **`null`/`undefined`**, wrong types, plain v2 **`SnapKeyring`**, and objects missing **`unwrap`** return false. Unit tests cover these cases; the unreleased changelog documents the export.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a0f000a01af618deb8e73c9559df9a56847e8d1a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/keyring-snap-bridge/src/v2/SnapKeyringV1Adapter.ts
+++ b/packages/keyring-snap-bridge/src/v2/SnapKeyringV1Adapter.ts
@@ -22,7 +22,7 @@ export function isSnapKeyringV1Adapter(
 
   const adapter = keyring as { type?: KeyringType; unwrap?: () => Keyring };
 
-  return adapter.type === KeyringType.Snap && adapter.unwrap?.() !== undefined;
+  return adapter.type === KeyringType.Snap && typeof adapter.unwrap === 'function';
 }
 
 /**
--- a/packages/keyring-snap-bridge/src/v2/SnapKeyringV1Adapter.ts
+++ b/packages/keyring-snap-bridge/src/v2/SnapKeyringV1Adapter.ts
@@ -22,7 +22,9 @@ export function isSnapKeyringV1Adapter(
 
   const adapter = keyring as { type?: KeyringType; unwrap?: () => Keyring };
 
-  return adapter.type === KeyringType.Snap && typeof adapter.unwrap === 'function';
+  return (
+    adapter.type === KeyringType.Snap && typeof adapter.unwrap === 'function'
+  );
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/keyring-snap-bridge/src/v2/SnapKeyringV1Adapter.ts
```diff
@@ -1,7 +1,30 @@
...
+  //
+  // This avoids relying on `instanceof` checks, which can fail in certain module resolution scenarios (e.g.
+  // when multiple versions of the same class exist).
+  if (keyring === null || keyring === undefined) {
+    return false;
+  }
+
+  const adapter = keyring as { type?: KeyringType; unwrap?: () => Keyring };
+
+  return adapter.type === KeyringType.Snap && adapter.unwrap?.() !== undefined;
```
mathieuartu: Is there a possible, more future-proof discriminant between V1 and V2 Snap Keyrings?

ccharly: Well, unfortunately, the `SnapKeyringV2` "as v1" (adapted) uses the same `KeyringType` than v2 instances 😬 (mostly cause we cannot re-use the old legacy Snap keyring one, as the semantic was different: legacy -> 1:N, v2 -> 1:1).

For now, the `unwrap` should do the trick.

I've added a new test to make sure that using a v2 instance won't return `true` with this function

ccharly: https://github.com/MetaMask/accounts/pull/618/commits/6be22a64289e4a08fe9eabdd620a3b52498cd7d6

mathieuartu (approved):

PR #624, opened 2026-08-19
fix: eip712 domains not being valid for gen 5 devices

montelaidev:
**Description**

EIP-712 signing fails on Ledger Gen5 devices with "the signature doesn't match the right address" when the payload omits the EIP712Domain type (common — ethers-style dapps infer it from the domain object keys).
The missing type gets defaulted to an empty list, so both the device and local verification hash an empty domain struct instead of the real domain values. Gen5 devices handle this undefined case differently than older devices, causing the signature check to fail.

Fix: derive the EIP712Domain type from the domain object keys before signing and verification. Payloads that already declare it are untouched — zero behavior change for well-formed input.

---

> [!NOTE]
> **Medium Risk**
> Changes EIP-712 normalization and signature recovery for payloads missing EIP712Domain; well-formed payloads are untouched, but incorrect derivation could affect permit/NFT signing on hardware wallets.
> 
> **Overview**
> Fixes **Ledger Gen5** failures when dapps send EIP-712 payloads with a populated `domain` but no `types.EIP712Domain` (common ethers-style shape). Without this, `@metamask/eth-sig-util` sanitization treats the domain type as empty, so the host hashes a zero-field domain while the device behaves differently—leading to **“The signature doesn't match the right address”**.
> 
> Adds **`withDerivedEip712Domain`**, which builds `EIP712Domain` from canonical domain keys (in contract order, skipping null/undefined and non-standard keys) when the type is missing or empty; payloads that already declare it are returned unchanged. **`LedgerKeyring.signTypedData`** runs this normalization before `sanitizeData`, sends the derived types to the device, and verifies with the same normalized data. The helper is **re-exported** from the package entry; unit and keyring tests cover NFT permit-style payloads and on-chain domain separator alignment.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c99a1a16d2b3f862c3202f89193b431d7936d995. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/keyring-eth-ledger-bridge/CHANGELOG.md
+++ b/packages/keyring-eth-ledger-bridge/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fix Ledger Gen5 EIP-712 signing for payloads that omit the `EIP712Domain` type by deriving it from the `domain` object keys before sending to the device and before signature verification ([#TODO](https://github.com/MetaMask/accounts/pull/TODO))
+- Fix Ledger Gen5 EIP-712 signing for payloads that omit the `EIP712Domain` type by deriving it from the `domain` object keys before sending to the device and before signature verification ([#624](https://github.com/MetaMask/accounts/pull/624))
 
 ## [13.0.2]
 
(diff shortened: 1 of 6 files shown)


montelaidev:
@metamaskbot publish-preview

hmalik88 (approved):

PR #440, opened 2026-01-19
feat: add MPC keyring

matthiasgeihs:
Add MPC Keyring

PR #627, opened 2026-09-02
feat: add Ethereum MPC keyring

matthiasgeihs:
> **Do not merge yet:** this PR still vendors MFA wallet packages as local `.tgz` files under `packages/keyring-eth-mpc/`. Replace those tarballs with published npm dependencies before merging.

## Summary
- Add `@metamask/eth-mpc-keyring`: a 2-party client/server MPC keyring with DKLS23 TSS, cloud backup sync, and vendored MFA wallet libs
- Support create/import init, key-share rotate/check/sync, and signing (transactions, personal_sign, typed data, EIP-7702 authorizations)
- Replaces the earlier WIP in #440 with a cleaned history rebased on `main`

## Test plan
- [ ] `yarn workspace @metamask/eth-mpc-keyring test`
- [ ] `yarn workspace @metamask/eth-mpc-keyring build`
- [ ] Smoke-check create/import and signing against a staging cloud/relayer if available

Made with [Cursor](https://cursor.com)


matthiasgeihs:
Replaces https://github.com/MetaMask/accounts/pull/440 with a cleaned history rebased onto `main`.

packages/keyring-eth-mpc/src/mpc-keyring.ts
```diff
@@ -0,0 +1,707 @@
...
+      token,
+      backupId,
+      encryptedKeyShare: await this.#encryptKeyShare(keyShare),
+    });
+
+    this.#applyKeyState({
+      ...state,
+      keyShare,
+      backupId,
+    });
```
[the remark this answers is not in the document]
matthiasgeihs: fixed in cb37f580af1072bfceea36aab893d26e033f6f39

packages/keyring-eth-mpc/src/mpc-keyring.ts
```diff
@@ -0,0 +1,707 @@
...
+            key: keyShare,
+            signers: bindings,
+            message: hash,
+            networkSession: netSession.createSubsession('tss-sign'),
+            setup: tssSetup,
+          });
+          return toEthSig(signature, hash, keyShare.publicKey);
+        } catch (error) {
+          this.#applyKeyState({ ...state, tssSetup: null });
+          throw error;
```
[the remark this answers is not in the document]
matthiasgeihs: resolved in e5a53ee5dbd2561e773d806583b044118f906318

matthiasgeihs:
**Do not merge yet:** this PR still vendors MFA wallet packages as local `.tgz` files under `packages/keyring-eth-mpc/`. It should not be merged until those tarballs are replaced with published npm dependencies.

packages/keyring-eth-mpc/src/mpc-keyring.ts
```diff
@@ -0,0 +1,782 @@
...
+        token,
+        epoch: nextEpoch,
+      });
+
+      this.#applyKeyState({
+        ...state,
+        keyShare,
+        shareEpoch: nextEpoch,
+        tssSetup: null,
+      });
```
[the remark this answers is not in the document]
matthiasgeihs: By design: crash before setActiveEpoch leaves the old epoch signable; recovery is to retry create/rotate (backend overwrites the pending epoch). syncKeyShare only reloads the active epoch on purpose. A short-circuit activate-when-ready path would avoid a redundant DKG, but isn’t required for correctness — follow-up, not a merge blocker.

packages/keyring-eth-mpc/src/types.ts
```diff
@@ -0,0 +1,57 @@
...
+import type { Json } from '@metamask/utils';
+
+export type ProfileTokenOpts = {
+  twoFactor?: boolean;
+  challenge?: Uint8Array;
+};
+
+export type Dkls23Lib = ConstructorParameters<typeof Dkls23TssLib>[0];
+
+export type MPCKeyringOpts = {
```
ccharly, line 13: Nit: Could you rename this to `MpcKeyringOpts` instead? Similarly to `Dkls23Lib`? This comment apply to any `MPC*` type here too!

packages/keyring-eth-mpc/src/util.ts
```diff
@@ -0,0 +1,344 @@
...
+    'AES-GCM',
+    false,
+    ['encrypt'],
+  );
+  const ciphertext = await globalThis.crypto.subtle.encrypt(
+    { name: 'AES-GCM', iv },
+    cryptoKey,
+    plaintext,
+  );
+  return concatBytes(iv, new Uint8Array(ciphertext));
```
ccharly, line 68: <s>
We have a pattern when relying on crypto, where we usually pass along the an object/config that contains the cryptographic functions we need.

This way, we can override them with native support on the clients if needed. That's mostly relevant for mobile which can highly benefit from that.

You can find an example in `key-tree` here:
- https://github.com/MetaMask/key-tree/blob/eb48f14a9effd9736b059a09b3bbb28d268f0006/src/cryptography.ts#L50-L72

And we can also keep the `crypto.subtle` as a fallback.

I think that'd be nice if we were adding this support. WDYT?
</s>

I did discuss this internally, and actually we should now have an up-to-date package on mobile that should implement all the (native) crypto we need for this:
- https://github.com/MetaMask/metamask-mobile/pull/30716

Though, we might need to update the shims there to make sure `importKey` and `encrypt` are correctly wired.

packages/keyring-eth-mpc/src/util.ts
```diff
@@ -0,0 +1,344 @@
...
+    key,
+    'AES-GCM',
+    false,
+    ['decrypt'],
+  );
+  const plaintext = await globalThis.crypto.subtle.decrypt(
+    { name: 'AES-GCM', iv },
+    cryptoKey,
+    ciphertext,
+  );
```
ccharly: Same remark here for native crypto support.

packages/keyring-eth-mpc/src/util.ts
```diff
@@ -0,0 +1,344 @@
...
+
+/**
+ * Generate a session nonce: random bytes from the RNG, hex-encoded.
+ *
+ * @param rng - The random number generator.
+ * @returns Hex-encoded 32-byte random nonce.
+ */
+export function generateSessionNonce(rng: RandomNumberGenerator): Hex {
+  return bytesToHex(rng.generateRandomBytes(SESSION_NONCE_BYTE_LENGTH));
+}
```
ccharly, line 34: I think we could move all the "crypto" implementation in a separate module, ideally a `cryptographic.ts` would be nice. WDYT?

packages/keyring-eth-mpc/src/cloud.ts
```diff
@@ -0,0 +1,312 @@
...
+  token: string,
+  errorPrefix: string,
+): Promise<Result> {
+  const response = await fetch(url, {
+    method: 'GET',
+    headers: {
+      Authorization: `Bearer ${token}`,
+    },
+  });
+  return parseJsonResponse(response, errorPrefix);
```
ccharly, line 56: If somehow the remote end is not available or whatever, we won't be retrying the request here. Do we want to add more resilience there?

packages/keyring-eth-mpc/src/cloud.ts
```diff
@@ -0,0 +1,312 @@
...
+): Promise<Result> {
+  if (!response.ok) {
+    throw new Error(`${errorPrefix}: ${response.statusText}`);
+  }
+
+  const text = await response.text();
+  if (text.length === 0) {
+    return undefined as Result;
+  }
+  return JSON.parse(text) as Result;
```
ccharly, line 34: We own the backend here, so we can probably make this assumption, but I wonder if adding response schema validation would make the whole thing slightly better.

If somehow the backend schema's drift from ours, that would lead to runtime errors potentially.

Like we could re-use `@metamask/superstruct` most probably? We have some new decorators like `sensitive` to wrap field validation and redact the values being tested.

But we could also just use `if (!is(response, *Response)) { throw new Error("bad response format"); }` to avoid leaking anything out of the response.

packages/keyring-eth-mpc/src/cloud.ts
```diff
@@ -0,0 +1,312 @@
...
+ * @param opts.baseURL - MPC backend base URL.
+ * @param opts.token - Profile token.
+ * @returns The server network id.
+ */
+export async function getNetId(opts: {
+  baseURL: string;
+  token: string;
+}): Promise<PartyId> {
+  const data = await postJson<{ netId: string }>(
+    `${opts.baseURL}/net-id`,
```
ccharly, line 98: Do we want to hardcode the `/v2` prefix of the backend used here. If ever we have a new `/v3` version on the backend, the keyring would also need to adapt, so I feel like this is more explicit to have it hardcoded here too, WDYT?

packages/keyring-eth-mpc/src/cloud.ts
ccharly: What about timeout handling also? If somehow the backend never respond and that we wrap a keyring call with a `:withKeyringV2` (on the `KeyringController`), we could hold the mutex forever, thus, blocking ALL keyring operations within MetaMask (until we hard restart the client).

packages/keyring-eth-mpc/src/mpc-keyring.ts
```diff
@@ -0,0 +1,784 @@
...
+      return;
+    }
+
+    const setup = this.#parseSetupParams(stateObj);
+    if (setup) {
+      this.#state = {
+        status: 'uninitialized',
+        setup,
+      };
+    }
```
ccharly, line 243: Nit: We could probably use `@metamask/superstruct` here too. There could be a way to wire the custom serializers too to a specific `struct` (refiner etc..).

packages/keyring-eth-mpc/src/mpc-keyring.ts
```diff
@@ -0,0 +1,795 @@
...
+          custodians,
+          networkSession: netSession.createSubsession('dkg-rotate'),
+        });
+      } finally {
+        await netSession.disconnect();
+      }
+
+      await storeKeyShareBackup({
+        baseURL: this.#cloudURL,
+        token,
```
ccharly, line 316: Edge-case: The token could get invalidated mid-session, I wonder if we should just just re-fetch it every time we need it?

packages/keyring-eth-mpc/src/mpc-keyring.ts
```diff
@@ -0,0 +1,795 @@
...
+        : undefined);
+    if (resolvedMode === undefined) {
+      return;
+    }
+
+    if (resolvedMode === 'create') {
+      await this.#setupCreate();
+    } else {
+      await this.#setupImport();
+    }
```
ccharly, line 270: Nit: We could probably use `serializeOp` here too?

`init` should be called very early in the keyring lifecycle, but for completeness, we could just call it here too? Or we can also use it in `#setup*` too since the rest of the code here is synchronous.

packages/keyring-eth-mpc/src/mpc-keyring.ts
```diff
@@ -0,0 +1,795 @@
...
+
+      const token = await this.#getProfileToken({ twoFactor: true });
+      const nonce = generateSessionNonce(this.#rng);
+      await startRotateKeyShares({
+        baseURL: this.#cloudURL,
+        token,
+        clientNetId: netCreds.partyId,
+        nonce,
+        expectedActiveEpoch: shareEpoch,
+      });
```
ccharly, line 295: Nit (refactor): Most of the network calls always forwardss the `baseURL` + `token`. I wonder if we should make the `cloud.ts` implementation a bit more "stateful" and construct an objects that would hold the URL + the `#getProfileToken` function. So we can use them automatically when forwarding the actual required parameters.

This way we could make potential back-off retry by refetching the token if needed, etc...

packages/keyring-eth-mpc/src/mpc-keyring.ts
```diff
@@ -0,0 +1,795 @@
...
+          const { signature } = await this.#tss.sign({
+            key: keyShare,
+            signers: bindings,
+            message: hash,
+            networkSession: netSession.createSubsession('tss-sign'),
+            setup: tssSetup,
+          });
+          return toEthSig(signature, hash, keyShare.publicKey);
+        } catch (error) {
+          this.#applyKeyState({ ...state, tssSetup: null });
```
ccharly, line 679: Maybe worth adding a small note that ANY errors would invalidate the TSS setup, meaning we would have to go through again on the next `#signHash`.

I don't recall the exact tradeoff here, but I think that's by design, right?

Also, a slight improvement could be to dissociated "unexpected errors" from other kind of errors (e.g. transport/network) so we can avoid invalidating the TSS setup when we know it's not necessary? But again, that's maybe by design, so we don't need this level of granularity.

packages/keyring-eth-mpc/src/mpc-keyring.ts
```diff
@@ -0,0 +1,795 @@
...
+      peerNetId,
+      TSS_HAVE_SETUP_MESSAGE_TYPE,
+    );
+    const peerPayload = JSON.parse(new TextDecoder().decode(peerBytes)) as {
+      haveSetup?: unknown;
+    };
+    const peerHaveSetup = peerPayload.haveSetup === true;
+    if (haveSetup && peerHaveSetup) {
+      return storedSetup;
+    }
```
ccharly, line 710: Nit: I guess we could skip asking the backend here if we `haveSetup` is `false` already? We'll have to run it anyway since the client does not have it?

packages/keyring-eth-mpc/src/mpc-keyring.ts
ccharly, line 1: Would be nice to add some JSdocs to most of those functions. The names are usually great and pretty self-describing, but I feel like it would not hurt to match them to some of the descriptions we already have in the docs. WDYT?

packages/keyring-eth-mpc/src/mpc-keyring.ts
```diff
@@ -0,0 +1,795 @@
...
+    let tssSetup: Uint8Array;
+    try {
+      const custodians = partyNetIds(netCreds.partyId, serverNetId);
+      const bindings = shareBindings(netCreds.partyId, serverNetId);
+      const createKeySession = netSession.createSubsession('dkg-create');
+      const tssSetupSession = netSession.createSubsession('tss-setup');
+      [keyShare, tssSetup] = await Promise.all([
+        this.#dkm.createKey({
+          custodians,
+          threshold: 2,
```
ccharly, line 556: That could be a constant, similar to the other ones we have for shares

packages/keyring-eth-mpc/src/cloud.ts
matthiasgeihs: I don't have a lot of experience with what are good practices for retry logic and timeouts.
Do we have MetaMask best practices for this?
What timeout length would you suggest?
What retry logic?

packages/keyring-eth-mpc/src/cloud.ts
matthiasgeihs: Thanks for the suggestion. Probably a good idea. Will explore further and come up with some proposal or come back with more questions.

packages/keyring-eth-mpc/src/cloud.ts
matthiasgeihs: The v2 path could also be part of the base URL. Or would you say there is a benefit in having it separate?

packages/keyring-eth-mpc/src/mpc-keyring.ts
matthiasgeihs: It depends whether this will trigger MFA each time. Not completely figured out yet. Token request interface will also probably need to be updated. Just a stub currently.

PR #628, opened 2026-09-03
feat: use bearer token in requests

ieow:
## Summary

- MPC cloud HTTP calls now send the profile token in an `Authorization: Bearer <token>` header instead of including `token` in JSON request bodies.
- The shared `postJson` helper accepts the token as a separate argument and applies it consistently across all MPC backend requests.
- Token-only endpoints (`getNetId`, `checkKeyShareBackupId`, `loadKeyShareBackup`) still use POST but now send an empty JSON body (`{}`).

## Test plan

- [x] Run `packages/keyring-eth-mpc` unit tests (`cloud.test.ts`)
- [x] Verify `getNetId` requests include the Bearer header and send `{}` as the body
- [x] Verify sign and backup helpers include the Bearer header and omit `token` from the JSON payload
- [ ] Confirm MPC backend accepts Bearer authentication for affected endpoints before merge

### DIFF
--- a/packages/keyring-eth-mpc/src/cloud.test.ts
+++ b/packages/keyring-eth-mpc/src/cloud.test.ts
@@ -50,7 +50,11 @@ describe('cloud helpers', () => {
       'https://cloud.example/net-id',
       expect.objectContaining({
         method: 'POST',
-        headers: { 'Content-Type': 'application/json' },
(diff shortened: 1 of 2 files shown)


ieow:
close in favor of #630

